PR #16, opened 2022-01-29
Feature: Sveltekit initialization

### LINKED ISSUE #2: Initialize a SvelteKit template
Blocked by #1 

Create a SvelteKit boilerplate in the `app` project with TypeScript. No Prettier, but with ESLint and `svelte-check`.

**Acceptance criteria**:
- `pnpm dev` in the `app` project starts a SvelteKit application
- `pnpm lint` in the `app` project runs ESLint on Svelte and TS and then `svelte-check` on Svelte
- `pnpm fix` in the `app` project runs ESLint on Svelte and TS in fixing mode

VanishMax:
Closes #2.

This template is created by ['create-svelte'](https://kit.svelte.dev/docs#introduction-getting-started) CLI.

@illright I believe that setting the Feature Sliced methodology and a project-wide Prettier is on you, right?


illright:
There's nothing to set up in Feature Sliced, really, it's only a matter of placing future components in the right places

.editorconfig
```diff
@@ -0,0 +1,8 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = false
```
illright, line 8: question: Why that? Those newlines are good, in my opinion

illright (approved):

.editorconfig
VanishMax: Oups, agree with you. Copied the config from one of the projects, and here's the result

PR #18, opened 2022-01-29
Feature: Presentation initialization

### LINKED ISSUE #3: Initialize a sli.dev template
Blocked by #1

Create a sli.dev boilerplate in the `presentation` project.

**Acceptance criteria**:
- `pnpm dev` in the `presentation` project starts sli.dev in dev mode

### LINKED ISSUE #14: Set up auto-deploying of the presentation
Blocked by #6 

This should run on each commit to `main` that made changes to any of the files in the `presentation` project. It should run after the CI for the `main` branch successfully passes. The deployment should be made to `/presentation`.

**Acceptance criteria**:
- A URL to the presentation should be added to the README.
- A presentation should be accessible at that URL.

VanishMax:
Closes #3
Closes #14 (Couldn't resist to steal the task from @alkaitagi)

Creates the template for the Sli.dev presentation and adds a GitHub action to publish it to Pages

### DIFF
--- a/.github/workflows/presentation.yaml
+++ b/.github/workflows/presentation.yaml
@@ -1,0 +1,31 @@
+name: Presentation publish
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - presentation/**
+
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v2.3.1
+
+      - name: Use pnpm
+        uses: pnpm/action-setup@v2.0.1
+        with:
+          version: 6.28.0
+
+      - name: Install and Build 🔧
+        run: |
+          pnpm install
+          pnpm build:presentation
+
+      - name: Deploy to Pages 🚀
+        uses: JamesIves/github-pages-deploy-action@4.1.4
+        with:
+          branch: gh-pages
+          folder: presentation/dist
--- a/package.json
+++ b/package.json
@@ -4,7 +4,9 @@
   "description": "A monorepo for the Cast app",
   "private": true,
   "scripts": {
-    "preinstall": "npx only-allow pnpm"
+    "preinstall": "npx only-allow pnpm",
+    "build:presentation": "cd presentation && pnpm build",
+    "build:app": "cd app && pnpm build"
   },
   "repository": {
     "type": "git",
--- a/presentation/package.json
+++ b/presentation/package.json
@@ -4,6 +4,13 @@
   "description": "Presentations of Cast projects made with sli.dev",
   "private": true,
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "dev": "slidev --open",
+    "build": "slidev build --base /cast/",
+    "export": "slidev export"
+  },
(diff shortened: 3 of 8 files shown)


.github/workflows/presentation.yaml
```diff
@@ -0,0 +1,31 @@
...
+        uses: pnpm/action-setup@v2.0.1
+        with:
+          version: 6.28.0
+
+      - name: Install and Build 🔧
+        run: |
+          pnpm install
+          pnpm build:presentation
+
+      - name: Deploy to Pages 🚀
```
illright, line 27: question: Why not to Cloudflare Pages? That's where we wanted to deploy the app, maybe we could deploy the presentation there too?

package.json
```diff
@@ -4,7 +4,9 @@
   "description": "A monorepo for the Cast app",
   "private": true,
   "scripts": {
-    "preinstall": "npx only-allow pnpm"
+    "preinstall": "npx only-allow pnpm",
+    "build:presentation": "cd presentation && pnpm build",
```
illright, line 8: suggestion (non-blocking): `pnpm` has a `--dir` option to run commands in a particular directory. The following might be more elegant:
```suggestion
    "build:presentation": "pnpm --dir presentation build",
```

But check that it works, because I'm not sure :D

illright (commented):
This kind of task-stealing might confuse the original assignee of the task and we'll end up duplicating efforts. In the future, could you please ask the assignee in the chat if they're willing to give up their task?

presentation/package.json
```diff
@@ -4,6 +4,13 @@
   "description": "Presentations of Cast projects made with sli.dev",
   "private": true,
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "dev": "slidev --open",
+    "build": "slidev build --base /cast/",
```
aabounegm, line 8: In relation to Lev's comment, would `/cast/` still be the base URL if we use CloudFlare Pages instead of GitHub's?

illright: No, `/presentation`

VanishMax: We could leave /presentation be in GitHub Pages, while the main app is in CloudFlare, since they are not related logically

aabounegm: No, we cannot because the repo's name is **cast**, so it will be `/cast/presentation`. The URL https://aabounegm.github.io/presentation would refer to the (non-existing) repo called **presentation**.

khaledismaeel (dismissed):
This approval is just for the sake of not blocking the PR; I'll blindly trust you guys with the design decisions this time.

presentation/package.json
illright: issue (blocking): this should be `/cast/presentation` then, as @aabounegm mentioned above

illright (changes requested):

presentation/package.json
illright: Or actually, no, never mind, it shouldn't

illright (approved):

presentation/package.json
aabounegm: why? Since the main app will be on CloudFlare Pages, why not leave it as just `/cast/` here?

PR #20, opened 2022-01-29
Feature: configure unit testing

### LINKED ISSUE #9: Configure unit testing in `app`
Blocked by #2 

Use the Svelte Testing Library with Jest.

**Acceptance criteria**:
- `pnpm test:unit` in the `app` project runs the test suite

khaledismaeel:
Resolves #9.

### DIFF
--- a/.github/workflows/unit-test-pr.yml
+++ b/.github/workflows/unit-test-pr.yml
@@ -1,0 +1,24 @@
+
+name: Run unit tests on Pull Requests
+
+on:
+  pull_request:
+    paths:
+      - app/**
+
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v2.3.1
+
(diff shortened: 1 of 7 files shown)


aabounegm:
Pair-programmed avec moi, but we still want feedback over where to place the tests (next to the component, or in `tests` or `__tests__` directory?), and also on consistency in the config files format (like, we `.babelrc`, but `jest.config.mjs`)

illright:
I think that tests should be placed in slices of functionality, i.e. one level below the components

Regarding config files, do we really need Babel? I think it wasn't necessary when I was configuring the testing of Apparitions

khaledismaeel:
I don't have concrete information on whether Babel is really needed here or not, but if I understood correctly, we need it for compatibility of TS and ES6 (we weren't sure what we're gonna be dealing with so we just decided to stay safe).

Do you suggest to remove it?

illright:
It's fine for now, maybe I'll look into removing it in the future

illright (approved):

Issue #6, opened 2022-01-28
Set up a CI checking workflow


illright:
Blocked by #9 and #10.

**Acceptance criteria**:
- every PR is validated with `pnpm run lint` in the workspace root and `pnpm run test:unit` and `pnpm run test:e2e` in the `app` project
- pnpm's dependencies are cached across workflow runs

PR #24, opened 2022-02-01
Set up Prettier

### LINKED ISSUE #12: Configure Prettier project-wide
Blocked by #2 

**Acceptance criteria**:
- `pnpm lint` in the workspace root checks all supported files with Prettier and then runs `lint` in `app`.
- `pnpm fix` in the workspace root tries to fix all formatting issues and then runs `fix` in `app`.

illright:
Closes #12

### DIFF
--- a/.github/workflows/unit-test-pr.yml
+++ b/.github/workflows/unit-test-pr.yml
@@ -1,4 +1,3 @@
-
 name: Run unit tests on Pull Requests
 
 on:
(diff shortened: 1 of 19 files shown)


aabounegm (approved):
It would be great if you add the prettier check to the CI as well.
And this is definitely a nitpick, but maybe add to VS Code's workspace extension recommendations and make it format on save? Or actually, I'm not sure if the extension is required since the Svelte extension uses it under the hood, but the setting might be useful.

Anyway, it's alright even without all that (and I can do them later if you don't want to)

illright:
The CI part has a separate task: #6

PR #33, opened 2022-02-04
Setup Husky and precommit hook

### LINKED ISSUE #13: Set up Husky and lint-staged
Blocked by #12 

This is to ensure that no code that violates static checkers can be committed.

**Acceptance criteria**:
- Formatting issues and ESLint problems in staged files should prevent Git from making commits
- Git hooks should be installed automatically upon running `pnpm i` (out of the box in Husky)

aabounegm:
Installs Husky and configures a pre-commit hook to check for linting.
Closes #13

Note:
- Windows users might need to set `npm config set script-shell powershell` since the `;` in the `lint` script doesn't work in the default `cmd` shell
- After this is merged, everyone needs to run `pnpm i` to install the hooks locally.


illright:
What about lint-staged?

aabounegm:
I didn't notice that it's required (it was only in the issue title, not description or acceptance criteria). And I think that linting everything, while a little more time-consuming, can help ensure the quality of the entire codebase at all times.

illright:
But that's a hook, we're still going to have CI to ensure code quality. I believe hooks should run as fast as they can

.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-pnpm lint
+pnpx --yes lint-staged
```
illright, line 4: issue (non-blocking): `pnpx` is deprecated, use `pnpm dlx` instead. But better yet, why not install it as a dev dependency and use `pnpm exec`? It's faster than trying to fetch it, even if from cache.

package.json
```diff
@@ -61,5 +61,9 @@
     "ts-jest": "^27.1.3",
     "tslib": "^2.3.1",
     "typescript": "^4.5.5"
+  },
+  "lint-staged": {
```
illright, line 65: suggestion (non-blocking): perhaps we could move it in a separate file [like this](https://github.com/illright/draw-ill-help/blob/main/.lintstagedrc.yaml)? In that example, there are also slightly better commands, you may wanna use them instead.

illright (approved):
Approved with non-blocking comments, feel free to ignore

.husky/pre-commit
aabounegm: I also wondered why the installation script (`npx mrm@2 lint-staged` from the docs) used `npx` instead of installing it, but I left it as it is (only changed `npx` to `pnpx`)

PR #35, opened 2022-02-04
Set up E2E testing

### LINKED ISSUE #10: Configure E2E testing in `app`
Blocked by #2.

Use Cypress.

**Acceptance criteria**:
- `pnpm test:e2e` in the `app` project runs the E2E test suite

aabounegm:
Configures End-to-End testing via Cypress and adds it to the PR checks.

Closes #10

_Note: Linux users should make sure to have the [required dependencies](https://docs.cypress.io/guides/getting-started/installing-cypress#Linux) first_

### DIFF
--- a/.github/workflows/unit-test-pr.yml
+++ b/.github/workflows/unit-test-pr.yml
@@ -2,7 +2,7 @@ name: Run tests on Pull Requests
 
 on:
   pull_request:
-    types: [opened, reopened]
+    types: [opened, synchronize, reopened]
 
 jobs:
   build-and-test:
--- a/.github/workflows/unit-test-pr.yml
+++ b/.github/workflows/unit-test-pr.yml
@@ -34,5 +34,8 @@ jobs:
       - name: Run Cypress E2E tests 🧪
         uses: cypress-io/github-action@v2
         with:
-          start: npm run dev
+          start: pnpm run dev
           wait-on: 'http://localhost:3000'
+          # we have already installed all dependencies above
+          install: false
+          cache-key: ${{ runner.os }}-node-${{ hashFiles('**/pnpm-lock.yaml') }}
--- a/.github/workflows/unit-test-pr.yml
+++ b/.github/workflows/unit-test-pr.yml
@@ -1,11 +1,11 @@
-name: Run unit tests on Pull Requests
+name: Run tests on Pull Requests
 
 on:
   pull_request:
     types: [opened, reopened]
 
 jobs:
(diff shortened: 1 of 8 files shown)


aabounegm:
The CI fail is because of the unit tests, not E2E tests. I tried fixing the type error locally, but nothing worked. The error is only reported when running in the CLI, but not in VS Code (IntelliSense)

illright:
What's up with the unit tests?

aabounegm:
As I said, I have no idea why it's not working, no matter what I add to the `types` field in `tsconfig` (tried `jest` and `@testing-library/svelte` and `@testing-library/jest-dom` and `ts-jest`, in all possible combinations), yet nothing worked other than removing the field altogether.

.github/workflows/unit-test-pr.yml
```diff
@@ -19,12 +19,23 @@ jobs:
...
+      - name: Check style and formatting 🧹
+        run: pnpm run lint
+
+      - name: Run unit tests 🧪
+        run: pnpm run test:unit
+
+      - name: Run Cypress E2E tests 🧪
+        uses: cypress-io/github-action@v2
+        with:
+          start: pnpm run dev
```
VanishMax, line 37: Testing in development environment might be different from the production one. I think, we could change this to build + start commands

cypress/integration/example.spec.ts
```diff
@@ -0,0 +1,9 @@
+/// <reference types="cypress" />
```
VanishMax, line 1: This reference will be used in all testing files, thus leading to code repetition. Cypress types could be added to `.tsconfig`, making them global

VanishMax (dismissed):

.github/workflows/unit-test-pr.yml
aabounegm: Good point :+1:

cypress/integration/example.spec.ts
aabounegm: I tried, but adding anything to the `types` field in `tsconfig` results in breaking the types for unit tests. Check the comment above

VanishMax: Anyway, we will be adding E2E tests into the Cypress folder, so what about a new `.tsconfig` in the `cypress` directory. It can extend the base config and add Cypress types only for this directory

aabounegm: Right, I forgot about that possibility. Done

illright (approved):

PR #36, opened 2022-02-05
Install TailwindCSS & Material Icons

alkaitagi:


illright (commented):
I suggest you to take a look at [`unplugin-icons`](https://github.com/antfu/unplugin-icons), a really great icon library for SvelteKit with all the icons we could ever want, yet with slim bundle sizes.

The available icons can be explored at https://icones.js.org/

alkaitagi:
> I suggest you to take a look at [`unplugin-icons`](https://github.com/antfu/unplugin-icons)

Sorry, couldn't get how to use the pack with Svelte, can't make the import from the `npm` package work.

illright (approved):
Alright, let's keep it at that, we can always switch later

Issue #26, opened 2022-02-01
Set up audio controls


illright:
This should probably be a single component, a feature in terms of Feature Sliced Design.

**Acceptance criteria**:
- [x] it allows pausing/resuming playback
- [x] it allows jumping forward 30s
- [x] it allows jumping backward 10s
- [x] it allows [scrubbing](https://en.wikipedia.org/wiki/Scrubbing_(audio))
- [x] it is integrated with the `<audio>` element that is in the layout (i.e. it doesn't add its own `<audio>` element)
- [ ] it is covered with unit tests

P.S. don't spend too much time styling, just use plain HTML

PR #37, opened 2022-02-05
Audio controls

alkaitagi:
@illright. @aabounegm, I need suggestions for refactoring. If I understand #26's requirement correctly, the component should accept `HTMLAudioElement` as a parameter. However, in my implementation, I rely upon direct bindings, and am unsure how to perform them on a passed element.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,18 +46,21 @@
     "@testing-library/svelte": "^3.0.3",
     "@typescript-eslint/eslint-plugin": "^5.10.1",
     "@typescript-eslint/parser": "^5.10.1",
+    "autoprefixer": "^10.4.2",
     "babel-jest": "^27.4.6",
     "eslint": "^7.32.0",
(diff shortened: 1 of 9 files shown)


aabounegm:
If I understand the requirement correctly, only TS code is required, and not an actual component. Think of it as a composable in Vue that accepts the `<audio>` element and returns a bunch of functions that can then be used in the component that adds the element to the layout.
Otherwise, it could mean that the component is to have only the controls, and will be passed the `HTMLAudioElement` as a prop, and it needs to update it manually (as in using the DOM API instead of Svelte's sugar). But again, I'm not entirely sure and let's wait for @illright's clarification

illright:
Also, just as a reminder to @alkaitagi — we agreed to stick to Git-style commit messages "Do X", "Add Y" as opposed to conventional commits

illright:
Superseded by #49

Issue #23, opened 2022-02-01
Set up podcast fetching on the frontend


illright:
Blocked by #10 

We should have an input field that you can enter a URL into (no validation) and that URL is fetched to return an audio file embedded with a native `<audio>` element with controls.

**Acceptance criteria**:
- The input is present on `/`
- The audio element on every route (that is, in the layout)
- The audio element is disabled when there's no music selected
- The interaction flow with the component is covered with e2e tests

Remember about Feature Sliced Design. I'm ready to discuss the architecture if you happen to have any doubts.

PR #44, opened 2022-02-09
Audio fetching

VanishMax:
Implements #23


VanishMax:
Hey @illright, please check the FS methodology in this code

src/routes/__layout.svelte
```diff
@@ -1,42 +1,7 @@
 <script lang="ts">
-  import '../app.css';
+  import { DefaultLayout } from 'src/lib/processes/layout';
```
illright, line 2: question: why did you introduce the `src` alias when you can import stuff from `$lib`?

src/lib/features/fetch-audio/ui/audio-fetch.svelte
```diff
@@ -0,0 +1,28 @@
+<script lang="ts">
+  import audioStore from '../model/audio';
+  
+  let url = '';
+  const submit = () => {
```
illright, line 5: issue: This function and the function below it belong to the model. Ideally, the `ui` segment should have as little logic as possible.

illright (changes requested):
1. Processes should not contain UI at all. The layouts are best positioned in the `routes` at app level.
2. The global styles should also remain at app level
3. I noticed an empty `lib/features/fetch-audio/model/types.ts` file, why is it needed?

~~P.S. The linter left some warnings. We should probably amend the command of linting in the CI to fail on warnings too.~~ never mind, I see the CI is failing as it should

P.S.S. This review is mostly about code adherence to FSD, I haven't checked if the code does what it should do yet.

src/routes/__layout.svelte
VanishMax: Oh, I just saw someone added this alias to the Vite config, and only after realized the useful $lib alias

src/lib/features/fetch-audio/ui/audio-fetch.svelte
VanishMax: I would argue about this. Isn't it a refactoring for the sake of refactoring?
I mean, isn't it is "as little as possible"?

illright: In this case, no, because it violates the rules of segments. FSD is flexible in the level of decomposition to prevent "refactoring for the sake of refactoring".

I would argue that any methodology at the start of the project feels like pointless refactoring because the project hasn't grown to be unmanageable yet. However, when (if) it does, refactoring would be much more painful, so I think it's reasonable to stick to clean architecture from the beginning.

VanishMax:
@illright @alkaitagi the PR is ready, waiting for your review.

But before doing it, let me explain why the tests look like this:
* I followed [Cypress best practices](https://docs.cypress.io/guides/references/best-practices) by using `data-cy` attributes instead of changing classes and ids
* The code of tests is organized with the use of the `support` folder that keeps constant data and custom commands. Think about it as the `features` slice in FS methodology
* I used the custom `ClientOnly` component with `<span data-cy="hydrated" />` in the layout to understand if the client-side has performed hydration of server-side generated code. The reason for this – no other logic (that I've found or @aabounegm proposed) worked to check hydration correctly.

VanishMax:
@illright actually, Cypress end-to-end tests do not need other libraries (because it can do anything, and others would need to know only Cypress), and i would propose using Testing library only for unit tests

illright:
> Cypress end-to-end tests do not need other libraries

That's true, but Testing Library isn't an additional framework like Jest. Testing Library is just an integrated set of queries that encourage writing tests that closely resemble how users use websites.

> because it can do anything

What I want to achieve with the Testing Library can be done in plain Cypress with `cy.get([role=button][name=something]')`, it's just quite a lot uglier.

> i would propose using Testing library only for unit tests

That said, I don't understand why you propose this

cypress/integration/home.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+  });
+
+  it('plays audio music by typing the link to an audio file', () => {
+    cy.get('[data-cy="audio-fetch-form"]').as('form').get('input').type(constants.audio_example);
+    cy.get('@form').get('button[type=submit]').click();
+    cy.expectPlayingAudio();
+  });
+
+  it('plays audio music by playing default sound', () => {
+    cy.get('[data-cy="audio-fetch-form"]').as('form').get('button[type=button]').click();
```
aabounegm, line 16: Why are you adding the `form` alias here but not using it later in the test?

cypress/integration/home.spec.ts
```diff
@@ -0,0 +1,19 @@
+import { constants } from '../support';
+
+describe('Home page', () => {
+  beforeEach(() => {
+    cy.visit('/');
+    cy.get('[data-cy="hydrated"]');
```
aabounegm, line 6: Are you gonna replace this with the solution suggested by Lev? The one at
https://gist.github.com/stevekinney/9fc5c5f69006722739546751df026ceb
I'd prefer it if you do

src/lib/widgets/audio/ui/audio.svelte
```diff
@@ -0,0 +1,42 @@
+<script lang="ts">
+  import { audioStore } from '$lib/features/fetch-audio';
+  import ScrubbingBar from './scrubbing-bar.svelte';
+
+  let audioEl = null as HTMLAudioElement | null;
```
aabounegm, line 5: I think you can leave it as 
```suggestion
  let audioEl: HTMLAudioElement;
```
to avoid the checks in the beginning of every function, since it will always be available whenever any of these functions is called. The only problem can happen when it used outside any function, but that should be clear straight away

src/lib/widgets/audio/ui/audio.svelte
```diff
@@ -0,0 +1,42 @@
...
+    currentTime = audioEl.currentTime || 0;
+    if (audioEl.buffered.length) {
+      bufferTime = audioEl.buffered.end(audioEl.buffered.length - 1) || 0;
+    }
+  };
+  
+  audioStore.subscribe((val) => {
+    if (!audioEl) return;
+    audioEl.pause();
+    console.log('here', val);
```
aabounegm, line 23: I think you forgot to remove this line after testing

src/lib/widgets/audio/ui/audio.svelte
```diff
@@ -0,0 +1,42 @@
...
+    audioEl.src = val;
+    audioEl.currentTime = 0;
+    audioEl.addEventListener('timeupdate', timeUpdate);
```
aabounegm, line 26: This feels a little error prone in the future. I understand that attaching the same event listener multiple times will discard the extra instances, but if the listener got changed to something like `() => timeUpdate(time)` or anything similar, then it will be called many times.
Also, why is it not just `on:timeupdate={timeUpdate}` on the element itself?

cypress/support/expect-playing-audio.ts
```diff
@@ -0,0 +1,13 @@
+const expectPlayingAudio = () => {
```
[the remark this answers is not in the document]
VanishMax: > This file runs before every single spec file

Yes, it runs. But in this context a "run" means registering Cypress custom commands, so they can be used as `cy.expectPlayingAudio()`

cypress.json
```diff
@@ -1,5 +1,9 @@
 {
   "baseUrl": "http://localhost:3000/",
   "componentFolder": "src",
-  "testFiles": "**/*spec.{js,ts}"
+  "testFiles": "**/*spec.{js,ts}",
+  "supportFile": "cypress/support/index.ts",
+  "pluginsFile": false,
+  "viewportHeight": 800,
+  "viewportWidth": 720
```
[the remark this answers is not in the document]
VanishMax: I did a mistake here, thanks for noticing

src/lib/widgets/audio/ui/audio.svelte
VanishMax: Thanks, your solution is neat

cypress/support/expect-playing-audio.ts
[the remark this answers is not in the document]
aabounegm: I understand that, but not every test needs this command. I don't know how much overhead this incurs, or if it is the usual thing to do in Cypress, so I just wanted to point it out.

illright (approved):

PR #49, opened 2022-02-15
Added podcasts screen & reworked audio control

alkaitagi:


src/app.css
```diff
@@ -1,8 +1,48 @@
-:root {
-  font-family: Arial, -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu,
-    Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
-}
+@import url('https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap');
```
illright, line 1: question (non-blocking): why do you import the font in CSS? I believe it is less performant since the CSS has to be loaded and parsed before the request for the font files is made.

src/audio-service.ts
```diff
@@ -0,0 +1,27 @@
+import { writable } from 'svelte/store';
```
illright, line 1: issue: This file looks like a good candidate for an `audio` entity. In terms of FSD, this code should be located in `src/entities/audio/model/audio-instance.ts`. The filename can be whatever, though.

src/features/audio-controls.svelte
```diff
@@ -1,61 +1,29 @@
 <script lang="ts">
```
illright, line 1: issue: in FSD, components can't be placed in `features/` directly. `features` is a _layer_, containing domain-specific _slices_ inside of it. This file could be placed at `src/features/playback-controls/ui/controls.svelte`

src/loader-service.ts
```diff
@@ -0,0 +1,35 @@
...
+  console.log('loading podcast', id);
+  return podcast;
+}
+
+export async function getEpisodes(id: string) {
+  console.log('loading episodes', id);
+  return episodes;
+}
+
+const podcast = {
```
illright, line 13: praise: thanks for adding sample data :)

src/audio-service.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const _audio = new Audio();
```
illright, line 3: issue: I'm getting an exception here:
```
ReferenceError: Audio is not defined
    at /home/illright/work/cast/src/audio-service.ts:3:19
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async instantiateModule (/home/illright/work/cast/node_modules/.pnpm/vite@2.7.13/node_modules/vite/dist/node/chunks/dep-f5552faa.js:60105:9)
```

illright (changes requested):
Haven't reviewed everything yet. If you need help understanding how to structure code with FSD, hit me up

src/app.css
alkaitagi: Didn't do it consciously, just copied to code from GFonts. I will move it to `<link>`.

src/audio-service.ts
alkaitagi: Should work now as I pass an `HTMLAudioElement` bound from the layout.

src/features/episode-presentation/ui/download-link.svelte
```diff
@@ -0,0 +1,17 @@
+<script lang="ts">
+  export let percentage: number | undefined = undefined;
+  export let onClickDownload: (() => void) | undefined = undefined;
```
illright, line 3: issue (non-blocking): that's kind of React-style to pass event handlers as props to components. 

suggestion (non-blocking): perhaps we could separate responsibilities a bit and extract the actual download link into a feature component and have that component forward the click event to the parent? The downloading progress would then be handled by another component.

The reason I think it's better is because this component's prop interface looks a bit odd to me, particularly the `percentage` prop. It's like this component implements some logic of the download, but not all of it. But I see the benefit of having it as a single component too, since it's mostly the same element in the UI.

src/features/episode-presentation/ui/episode-card.svelte
```diff
@@ -0,0 +1,54 @@
+<script lang="ts">
+  import type { PodcastEpisode } from 'src/podcast';
+  import DownloadLink from './download-link.svelte';
+
+  export let liked = false;
+  export let selected = false;
+  export let episode: PodcastEpisode;
+  export let downloadingPercentage: number | undefined = undefined;
+  export let onClickDownload: (() => void) | undefined = undefined;
```
illright, line 9: issue (non-blocking): same comment about React-style event handlers

src/features/playback-controls/ui/controls.svelte
```diff
@@ -0,0 +1,28 @@
+<script lang="ts">
+  import { play, pause, seek, move, audio } from 'src/entities/audio/model/audio-instance';
```
illright, line 2: suggestion: I encourage you to keep FSD layers in the `src/lib` directory. For example, `src/lib/entities/audio/...`. The reason is two-fold:
* That's what Max and I were doing in the other PRs
* That allows importing with the SvelteKit-native alias, `$lib`, without having to define our own aliases

issue: in terms of FSD, code outside of a slice is prohibited to go inside that slice's structure. Every slice should have an `index.ts` defining its public interface, and other slices should stop there:
```suggestion
  import { play, pause, seek, move, audio } from '$lib/entities/audio';
```

src/routes/__layout.svelte
```diff
@@ -1,42 +1,30 @@
...
+  <main class="relative margin-auto h-full w-full max-w-md p-4 bg-slate-900 text-white">
+    <slot />
 
-  @media (min-width: 480px) {
-    footer {
-      padding: 40px 0;
-    }
-  }
-</style>
+    <div
```
illright, line 16: praise: wow, you actually implemented the bottom bar. That was supposed to be in the upcoming sprints, but that's cool.

src/features/playback-controls/ui/controls.svelte
```diff
@@ -0,0 +1,28 @@
...
+  export let showSeeker = false;
+  $: percent = (100 * $audio.currentTime) / $audio.duration;
+</script>
+
+<div class="flex flex-col gap-5">
+  {#if showSeeker}
```
illright, line 8: question: are there cases where we don't want to show the scrubbing bar?

src/loader-service.ts
```diff
@@ -0,0 +1,35 @@
+import type { PodcastEpisode, PodcastSeries } from './podcast';
+
+export async function getPodcast(id: string) {
+  console.log('loading podcast', id);
```
illright, line 4: suggestion (non-blocking): perhaps we could leave a TODO comment here? I find that they nicely delineate the code that's done from the code that should be returned to.

src/loader-service.ts
```diff
@@ -0,0 +1,35 @@
+import type { PodcastEpisode, PodcastSeries } from './podcast';
```
illright, line 1: suggestion: this whole file looks like it belongs to `entities/episode/model`

src/routes/podcasts.svelte
```diff
@@ -0,0 +1,43 @@
+<script lang="ts">
```
illright, line 1: suggestion: I propose we treat `routes` as being on the `app` layer of FSD. That said, ideally we'd want to keep as little page-based responsibility here. Perhaps we could move all this code to `pages/podcast-episodes/ui`?

illright (changes requested):
I see you've created a feature called `episode-presentation`.  

issue: in terms of business logic of the app, I don't think that's a feature that brings benefit to the user. The features here are to play the podcast, to download and to like it. Episode presentation looks to me as merely a necessary step from developers to enable users to have these features.

suggestion: I would suggest spreading out the parts in that feature to the layers above and below, for example, like this:
* create a feature `download-episode` and place the download link component there
* create a feature `play-episode` and move the Play button there
* create a feature `like-episode` and move the Like button there
* move the markup of the episode card to `entities/episode/ui`, leave slots for buttons to play, download and like podcasts.
* compose the markup of the episode card with the feature buttons in `widgets/episode-card/ui`

You might think this is excessive decomposition, which is true, but in my experience I haven't encountered problems from following FSD _too much_, instead it made the components much smaller and more readable. Especially since we're using Tailwind, where styles end up being a huge class string, something that @VanishMax complained about once.

praise: it's a lot of work you've done here, I appreciate that.

aabounegm:
152 commits?! I think the base branch should be changed, at least have `main` merged into it

illright (approved):

PR #55, opened 2022-02-18
Hide parts of the README that are not ready yet

illright:


khaledismaeel:
I don't think there's much to be said here. Merging.

PR #67, opened 2022-02-23
Finish the podcast episodes screen, refine and style playing controls

### LINKED ISSUE #26: Set up audio controls

### LINKED ISSUE #40: Style the playing controls according to the design
Just the elements that are covered in #26. A design link is available in #27.

**Acceptance criteria**:
* The playing controls look identical to the proposed design

### LINKED ISSUE #43: Construct the Podcast Episodes screen
Blocked by #39. 

Implement the layout proposed in #29.

**Acceptance criteria**:
* The page and its elements should adhere to Feature Sliced Design
* The page should reside at `/podcasts/:podcast-id`
* ~~A link to this page should be available on the home page for every podcast fetched from Supabase~~ _outdated_
* The page should look identical to the proposed design

illright:
Sorry, this is a large pull request, doing almost everything at once. Contrary to acceptance criteria, I didn't cover the playing contorls with unit tests for two reasons:
1. I want to slightly refactor the architecture of the interaction with the `<audio>` element
2. I promised to make all my PRs until Thursday
3. Did I say two reasons? No, there's definitely more
4. These changes are blocking other tasks, so we gotta get them out there ASAP.

Notable changes:
* The icon buttons are accessible with names now
* The link buttons are keyboard-accessible now
* Install the Iconify Intellisense extension as per the recommendation, it's a godsend
* Duration and position in audio are now tracked in separate stores (which are readable-only so far, ideally we'd want to be able to write to the position)
* The scrubbing bar is styled as best as could be (we still have the buffered zones to add to it though)
* No more check for playing audio in the integration test (it doesn't work anymore, we don't have the `<audio>` element)

Closes #26, closes #40, closes #43


src/lib/features/playback-controls/model/toggle-global-playback.ts
```diff
@@ -3,6 +3,10 @@ import { play, pause, audio } from '$lib/entities/audio';
 
 export function toggleGlobalPlayback() {
   const $audio = get(audio);
+  if ($audio === undefined) {
+    return;
+  }
```
VanishMax, line 8: Question: why are these three lines of code everywhere? Is there no way to ensure the `$audio` is not `undefined`?

VanishMax (changes requested):
Todo: finish the style of audio controls according to #40, because they are currently no full-width on mobile screen

src/lib/features/playback-controls/model/toggle-global-playback.ts
illright: Well, that's the only way)

illright:
@VanishMax the audio controls are flexible, they will grow when we place them in the Now Playing widget. I don't think there's anything to change here

VanishMax (approved):
Talked with Lev privately. We discussed FSD and the design behind this PR. Approving

src/lib/shared/ui/format-duration.ts
```diff
@@ -14,5 +15,15 @@ export function formatDuration(durationInSeconds: number) {
...
+  let sliceStart = 11;
+  if (durationInSeconds < 10 * 60 * 60) {
+    sliceStart++;
+
+    if (durationInSeconds < 1 * 60 * 60) {
+      sliceStart += 2;
+    }
+  }
+
+  return new Date(durationInSeconds * 1000).toISOString().substring(sliceStart, 19);
```
aabounegm, line 28: To be honest, this feels like a really dirty workaround to avoid using `getHours`/`getMinutes`/`getSeconds`. At least you could've used RegEx to extract the time from the ISO string.
The solution I would prefer is to use `new Intl.DateTimeFormat('en-GB', { timeStyle: 'medium' }).format(date)` (from [MDN - `Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat))

src/lib/features/playback-controls/ui/scrubbing-bar.svelte
```diff
@@ -0,0 +1,161 @@
+<script lang="ts">
+  import { formatDuration } from '$lib/shared/ui';
+
+  import { createEventDispatcher } from 'svelte';
+
+  export let duration: number;
+  export let position: number;
+  // export let buffered: TimeRanges;
```
aabounegm, line 8: Is this commented-out code left here intentionally?

aabounegm (commented):
I realize the PR is already merged, but I had already started reviewing so I just completed it now 😁

src/lib/features/playback-controls/ui/scrubbing-bar.svelte
illright: Yes, I've been meaning to come back to it at some point to render the buffered ranges.

src/lib/shared/ui/format-duration.ts
illright: It's not ideal since it leaves the leading zero in the hours, that's not exactly something you'd want to see in a duration representation. I agree that this is hacky, but doing something with a bunch of `if` statements felt like overengineering to me. Idk how to improve this

src/lib/features/download-episode/ui/specs/download-link.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+jest.mock('~icons/ic/twotone-file-download-done', () => ({ default: null }), { virtual: true });
+
+import { useLocalVars } from '$lib/shared/lib/jest-hacks';
+import { DownloadLink } from '../..';
+
+it('forwards the click event on clicks', async () => {
+  const mockClickHandler = jest.fn();
+  const user = userEvent.setup();
+
+  const downloadLink = useLocalVars(svelte`<DownloadLink on:click={mockClickHandler} />`, [
```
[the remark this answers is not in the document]
illright: I thought about it, initially it was meant to be that way, but I decided against it for two reasons:
* This closely matches React's hooks with a dependency array
* It organizes these arguments neatly, separating them from the markup of the component

src/lib/entities/podcast/ui/podcast-display.svelte
```diff
@@ -0,0 +1,17 @@
...
+  export let podcast: Podcast;
+</script>
+
+<div class="flex flex-row gap-4 p-2">
+  <img class="w-24 h-24 rounded-xl" src={podcast.coverUrl} alt="" />
+  <div class="flex flex-col gap-1 text-xs pt-2">
+    <H1>{podcast.title}</H1>
+    <p>{podcast.author}</p>
+    <p>
+      {podcast.episodes.length} episode{s(podcast.episodes.length)} / <slot name="download-all" />
```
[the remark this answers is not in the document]
illright: According to Feature-Sliced Design, we cannot reference `features` from `entities`, and downloading is a feature (because it involves the user interacting with the UI). I already had a conversation with Max and Khaled, selling them my vision of FSD, perhaps you and I could have it as well :)

src/lib/entities/episode/ui/specs/episode-card-shell.spec.ts
```diff
@@ -0,0 +1,29 @@
+import svelte from 'svelte-inline-compile';
+import { render } from '@testing-library/svelte';
+
+import { useLocalVars } from '$lib/shared/lib/jest-hacks';
+import type { Episode } from '$lib/shared/api';
+import { EpisodeCardShell } from '../..';
```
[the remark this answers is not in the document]
illright: Maybe, but with this I also wanted to test the public interface of the slice, because that's an important part of usage. I want these tests to reflect the usage of the component as closely as possible to the real one.

src/lib/entities/audio/model/audio-instance.ts
```diff
@@ -1,9 +1,53 @@
...
 
-export const audio = writable<HTMLAudioElement>();
+export const audio = writable<HTMLAudioElement | undefined>(undefined);
+
+/** Tracks the duration of the audio stream that is currently loaded in the `audio` store. */
+export const audioDuration = derived<typeof audio, number>(
+  audio,
+  ($audio, set) => {
+    if ($audio === undefined) {
+      set(NaN);
```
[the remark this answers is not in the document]
illright: That's just what [MDN specifies](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/duration#value) as the value when no media data is available, figured we could make it compliant.

src/lib/entities/audio/model/audio-instance.ts
```diff
@@ -1,9 +1,53 @@
...
+      return;
+    }
+
+    function watchDuration(event: Event) {
+      set((event.target as HTMLAudioElement).duration);
+    }
+
+    $audio.addEventListener('durationchange', watchDuration);
+
+    return () => $audio?.removeEventListener('durationchange', watchDuration);
```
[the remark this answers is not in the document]
illright: Yeah, of course, we don't know when this callback will be called. I'm not sure if TypeScript will complain, but it should

src/lib/features/download-episode/ui/specs/download-link.spec.ts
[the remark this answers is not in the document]
aabounegm: I don't see how React's hooks are relevant in any way (it's not like we're declaring dependencies or anything) but fine, I agree with the second point.

src/lib/shared/ui/format-duration.ts
aabounegm: A solution that does the above (with `Intl`) and then checks if `duration.startsWith('00:')` and then either `duration.slice(3)` or even just `duration.replace(/^00:/, '')` without the check would be a lot less hacky than the `sliceStart` thing, and even be much shorter.

src/lib/entities/audio/model/audio-instance.ts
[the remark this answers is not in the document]
aabounegm: We know for sure that this callback will not be reached if `$audio` was `undefined`. It also threw me off a little that TS didn't complain, but after playing around with it, I realized that it's static analysis was a lot smarter; if it sees `$audio = undefined;` anywhere in the function, it will not make that assumption and will require `?.`, but since it is in a closure and guards against `undefined` and never assigns it, the `$audio` parameter can never refer to `undefined` inside this callback, even if the original `audio` store became `undefined`.
The same applies in the `audioPosition` derived store as well.

src/lib/entities/podcast/ui/podcast-display.svelte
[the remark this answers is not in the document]
aabounegm: I think that would be helpful indeed since the names of the slices do not really stick in my head at first, they don't feel very intuitive

src/lib/shared/ui/format-duration.ts
illright: But you also need to cut out the leading zero if the hour count is 1-9: `1:00:00`, not `01:00:00`

illright: And honestly, now that I think about it, we should have done it with minutes too. Okay, this entire function should be rewritten properly. I'll make an issue to do that (#72)

aabounegm: Then change the RegEx to `^[0:]+` and it should handle that case as well

PR #71, opened 2022-02-24
Badge fix

### LINKED ISSUE #61: Fix a badge in the README to reflect the state of `main`
Currently, the CI status badge in the README reflects the last run on a pull request, which is not what we want. 

To avoid duplication of configuration that runs tests, refer to the [Action Composition](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) guide.

alkaitagi:
Closes #61.


.github/actions/unit-tests-action/action.yml
```diff
@@ -0,0 +1,37 @@
+name: Run unit tests
```
illright, line 1: suggestion (non-blocking): Perhaps we should phrase it differently to reflect that it's not just unit tests but all code quality tools

.github/workflows/unit-test-main.yml
```diff
@@ -0,0 +1,13 @@
+name: Run tests on `main` push
```
illright, line 1: suggestion (non-blocking): Also, since the badge has the form "<name> | passing", perhaps it would be better to name this workflow just "Tests on `main` push"?

illright (approved):
Nice. I left these two comments, but feel free not to take action on them

PR #76, opened 2022-02-27
Add the Uptime button to the README

illright:
I don't know if it's needed since we have a badge that links to the same page. It still looks kind of nice


aabounegm (approved):
I also don't like the duplication, but it still looks good. I guess we can keep it

PR #77, opened 2022-02-28
Fix the executability of the pre-commit hook

### LINKED ISSUE #64: Add `svelte-check` to the pre-commit hook
Currently, our pre-commit hook checks for ESLint issues and formatting problems. However, TypeScript errors go through freely. Remedy that by adding a `pnpm exec svelte-check [...]` to the `.husky/pre-commit` file.

P.S. Also an embarrassing detail: our `pre-commit` hook doesn't even run currently because it's not marked as executable. Please, mark it as executable while you're at it.

illright:
Closes #64. I took the liberty of implementing this myself because it's an important step in battling code that doesn't pass CI.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,3 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 pnpm exec lint-staged
+pnpm exec svelte-check --tsconfig ./tsconfig.json --fail-on-warnings --use-new-transformation
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "test:e2e": "cypress run",
     "test:e2e-open": "cypress open",
     "test:unit": "jest src",
-    "lint": "eslint --ignore-path .gitignore . --cache; prettier --check . --plugin-search-dir=.; svelte-check --tsconfig ./tsconfig.json",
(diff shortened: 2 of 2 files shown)


.husky/pre-commit
```diff
@@ -2,3 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 pnpm exec lint-staged
+pnpm exec svelte-check --tsconfig ./tsconfig.json --fail-on-warnings --use-new-transformation
```
aabounegm, line 5: This feels like such an expensive/time-consuming thing to run on every commit, since it checks the entire repo every time. How about we move it to `pre-push` instead?

aabounegm (commented):
I may not be the best person to review this one since the permission change doesn't affect me on Windows. I request that someone else with Linux verify it

.husky/pre-commit
illright: It's catching TypeScript errors, I think it's important to run before every commit. Also, it's not that time-consuming.

VanishMax:
I made a TypeScript mistake in a `.svelte` file, committed the change, waited for just 7 seconds, and it gave me this error. Works well!

image

VanishMax (approved):

PR #82, opened 2022-03-03
Refactor the API communication to support fetching objects from Supabase

illright:


khaledismaeel (commented):
I read the conversation on Telegram and I'm still kind of confused how we had to settle with all those `null`'s flying around. Why can't we handle this on the database level?

illright:
In a nutshell — the OpenAPI schema is generated by the constraints on the database, for example, the `required` field is generated from the `NOT NULL` constraint on a table column. The problem here is that these constraints can not be set on _view_ columns, only on table columns, and we need a view in order to merge the data from the `private` schema that contains the actual podcast metadata and the `storage` schema that contains the audio files.

khaledismaeel:
I see, it all links back to Supabase's incredibly weird database setup. Max and I struggled with this . In this case I'd love to handle this issue as far away from Supabase as possible.

khaledismaeel (approved):

src/lib/shared/api/not-null.ts
```diff
@@ -0,0 +1,3 @@
+export function notNull<ArgumentType>(argument: ArgumentType | null) {
```
aabounegm, line 1: It would have helped with type inference and flow analysis if the function was declared to be a type guard by setting the appropriate return type. Check this example inspired by [the docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions):
```ts
function isDefined<T>(val: T): val is NonNullable<T> {
  return !(val === undefined || val === null);
}
```

PR #83, opened 2022-03-03
Refactor the duration formatting code

### LINKED ISSUE #72: Rewrite the `formatDuration` function to strip leading zeros
For now, it only strips leading zeros of the hour count, doing `1:00:00` instead of `01:00:00`, but it doesn't do it for minutes. We should have durations like `5:00`, not `05:00`.

illright:
Closes #72


aabounegm (approved):
Ok, this was actually not as bad as I thought)
Longer than 2 lines, sure, but still readable :P

PR #84, opened 2022-03-03
Configure ESLint rules for Cypress

illright:


aabounegm (approved):
Didn't test, but it looks fine

PR #85, opened 2022-03-03
Remove the unnecessary fields and add the schema

illright:
Component folder and test files are only applicable to Cypress Component Testing, and that doesn't support Svelte as of now


VanishMax (approved):
Looks allright, but I don't see the usefullness in this PR – removing a couple of harmless lines. It could be done in any other PR, no one would notice or complain

PR #87, opened 2022-03-03
Audio polish

### LINKED ISSUE #69: Bind the audio to a physical `<audio>` element on the page
Without it, we cannot detect playing audio with Cypress.

**Acceptance criteria**:
- An actual `<audio>` node is present in the global `__layout.svelte`
- The audio store is bound to that instance
- The assertion in the integration tests that checks for playing audio is restored
- The integration tests pass

### LINKED ISSUE #70: Move away from storing an HTMLAudioElement in the store to specific value stores
All of these should be implementable without much code by simply [`bind:`ing to the corresponding attributes](https://svelte.dev/tutorial/media-elements) on the `<audio>` tag, so blocked by #69.

**Acceptance criteria**:
- `entities/audio` exposes a writable store for `currentTime` that defaults to 0
- `entities/audio` exposes a readable store for `duration` that defaults to NaN
- `entities/audio` exposes a writable store for `playbackRate` that defaults to 1
- `entities/audio` exposes a writable store for `paused` that defaults to true
- There should no longer be a store that allows direct access to the `<audio>` DOM node
- These stores should be wired up to the controls in `features/playback-controls`

alkaitagi:
Closes #69, closes #70.

### DIFF
--- a/src/lib/entities/audio/index.ts
+++ b/src/lib/entities/audio/index.ts
@@ -1,5 +1,5 @@
 export {
-  setAudio,
+  src,
   currentTime,
   duration,
   playbackRate,
--- a/src/lib/entities/audio/index.ts
+++ b/src/lib/entities/audio/index.ts
@@ -1,7 +1,9 @@
 export {
-  audio,
-  audioPosition,
-  audioDuration,
+  setAudio,
+  currentTime,
+  duration,
+  playbackRate,
+  paused,
   move,
   pause,
   play,
--- a/src/lib/entities/audio/index.ts
+++ b/src/lib/entities/audio/index.ts
@@ -4,8 +4,10 @@ export {
   duration,
   playbackRate,
   paused,
+  buffered,
   move,
   pause,
   play,
   seek,
 } from './model/audio-instance';
+export type { SvelteTimeRanges } from './model/audio-instance';
--- a/src/lib/entities/audio/model/audio-instance.ts
+++ b/src/lib/entities/audio/model/audio-instance.ts
@@ -1,10 +1,14 @@
 import { writable } from 'svelte/store';
 
+// Because apparently Svelte thinks it knows better how the Web APIs should be.
+export type SvelteTimeRanges = Array<{ start: number; end: number }>;
+
 export const src = writable('');
 export const currentTime = writable(0);
 export const duration = writable(NaN);
 export const playbackRate = writable(1);
 export const paused = writable(false);
+export const buffered = writable<SvelteTimeRanges>([]);
 
 export const play = (_src?: string) => {
   if (_src) src.set(_src);
--- a/src/lib/entities/audio/model/audio-instance.ts
+++ b/src/lib/entities/audio/model/audio-instance.ts
@@ -1,93 +1,24 @@
-import { writable, derived } from 'svelte/store';
+import { writable } from 'svelte/store';
 
-export const audio = writable<HTMLAudioElement | undefined>(undefined);
+let audio: HTMLAudioElement | undefined;
+export const currentTime = writable(0);
+export const duration = writable(NaN);
+export const playbackRate = writable(1);
+export const paused = writable(false);
 
-/** Tracks the duration of the audio stream that is currently loaded in the `audio` store. */
-export const audioDuration = derived<typeof audio, number>(
-  audio,
-  ($audio, set) => {
-    if ($audio === undefined) {
-      set(NaN);
-      return;
-    }
-
-    function watchDuration(event: Event) {
-      set((event.target as HTMLAudioElement).duration);
-    }
-
-    $audio.addEventListener('durationchange', watchDuration);
-
-    return () => $audio.removeEventListener('durationchange', watchDuration);
-  },
-  NaN
-);
-
-/** Tracks the playback position in the audio stream that is currently loaded in the `audio` store. */
-export const audioPosition = derived<typeof audio, number>(
-  audio,
-  ($audio, set) => {
-    if ($audio === undefined) {
-      set(0);
-      return;
-    }
-
-    function watchProgress(event: Event) {
-      set((event.target as HTMLAudioElement).currentTime);
-    }
-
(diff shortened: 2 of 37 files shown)


src/routes/__layout.svelte
```diff
@@ -2,20 +2,21 @@
...
 
+  let audioNode: HTMLAudioElement;
   onMount(() => {
-    audio.set(new Audio());
+    audioManager.set(audioNode);
   });
 </script>
 
-{#if $audio}
+{#if $audioManager}
```
illright, line 16: question: does this work? I have a feeling the `<audio>` node won't be mounted until the `audioManager` store gets a proper value, and it can only get a proper value when the `<audio>` node is mounted. Kind of a chicken-egg situation.

src/lib/entities/audio/model/audio-instance.ts
```diff
@@ -1,93 +1,24 @@
...
-  audio.update(($audio) => {
-    if ($audio === undefined) {
-      return;
-    }
-
-    $audio.currentTime += deltaInSeconds;
-    $audio.play();
-    return $audio;
-  });
+  if (audio) currentTime.set(audio.currentTime + deltaInSeconds);
```
illright, line 23: suggestion: Writable stores have an `update` method that receive the previous value of the store. This could help eliminate the dependency on `audio` here

src/lib/entities/audio/model/audio-instance.ts
```diff
@@ -1,93 +1,24 @@
...
-  audio.update(($audio) => {
-    if ($audio === undefined) {
-      return;
-    }
-
-    $audio.pause();
-    return $audio;
-  });
+export const play = (src?: string) => {
+  if (audio && src) audio.src = src;
```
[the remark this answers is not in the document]
illright: Well, `src` can't really be changed from within the `<audio>`, so we don't need to `bind:` it, we can simply pass it as a prop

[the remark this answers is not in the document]
alkaitagi: Then I don't understand how to decouple from the `audio` node if the only way of changing `src` is direct access to the node.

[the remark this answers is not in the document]
illright: Create a store that would contain the currently loaded `src` (defaulting to `undefined`). Then use that store in the `__layout.svelte` on the `<audio>` node like this:
```diff
<audio
- bind:this={audioNode}
  bind:duration={$duration}
  bind:currentTime={$currentTime}
  bind:paused={$paused}
  bind:playbackRate={$playbackRate}
+ src={$activeAudioUrl}
/>
```

We don't need direct access to the node that `bind:this` gives us, we only need the `<audio>` node to respond to changes in the `activeAudioUrl` store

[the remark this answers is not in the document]
alkaitagi: Of course, thank you very much. My understanding has lagged in the most important part.

illright (approved):
Now the audio code looks much cleaner. Nice!

src/lib/entities/audio/model/audio-instance.ts
```diff
@@ -1,115 +1,23 @@
-import { writable, derived } from 'svelte/store';
+import { writable } from 'svelte/store';
 
-export const audio = writable<HTMLAudioElement | undefined>(undefined);
+// Because apparently Svelte thinks it knows better how the Web APIs should be.
+export type SvelteTimeRanges = Array<{ start: number; end: number }>;
```
aabounegm, line 4: I don't understand why can't we use the built-in `TimeRanges` everywhere? It has a public constructor we can use for the default values (unlike `FileList`)

aabounegm, line 4: I don't understand why can't we use the built-in `TimeRanges` everywhere? It has a public constructor we can use for the default values (unlike `FileList`)

src/routes/__layout.svelte
```diff
@@ -1,28 +1,32 @@
...
   import IconVideoLibrary from '~icons/ic/twotone-video-library';
   import IconFolderShared from '~icons/ic/twotone-folder-shared';
+  import { src, duration, currentTime, paused, playbackRate, buffered } from '$lib/entities/audio';
   import { FullControls } from '$lib/features/playback-controls';
-  import { audio } from '$lib/entities/audio';
   import { IconButton } from '$lib/shared/ui';
 
-  import '$lib/app.css';
+  // Svelte can't handle its own type conversion with TypeScript
+  const bufferedNative = buffered as unknown as Writable<TimeRanges>;
```
aabounegm, line 10: But how does this work? The type `TimeRanges` is equivalent to 
```ts
type TimeRanges = {
  length: number;
  start: (i: number) => number;
  end: (i: number) => number;
};
```
while buffered has the type `Array<{start: number; end: number}>`

aabounegm, line 10: But how does this work? The type `TimeRanges` is equivalent to 
```ts
type TimeRanges = {
  length: number;
  start: (i: number) => number;
  end: (i: number) => number;
};
```
while buffered has the type `Array<{start: number; end: number}>`

aabounegm (commented):
I know my review was not requested and the PR is already merged, but I sometimes like to check the code anyway 😅

src/lib/entities/audio/index.ts
```diff
@@ -1,10 +1,13 @@
...
+  duration,
+  playbackRate,
+  paused,
+  buffered,
   move,
   pause,
   play,
   seek,
 } from './model/audio-instance';
+export type { SvelteTimeRanges } from './model/audio-instance';
```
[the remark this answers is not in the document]
illright: They can, I just didn't bother

src/lib/entities/audio/model/audio-instance.ts
illright: Not if we `bind:` we can't. Svelte does this conversion automatically, whether you want it or not

PR #89, opened 2022-03-03
Display buffered ranges on the scrubbing bar

### LINKED ISSUE #73: Highlight the buffered regions in the scrubbing bar
Get back to the commented out line in [features/playback-controls/ui/scrubbing-bar.svelte](https://github.com/aabounegm/cast/blob/main/src/lib/features/playback-controls/ui/scrubbing-bar.svelte)

illright:
Closes #73

Check out the Cloudflare demo, it actually works. Might want to open DevTools and right-click your reload button to reload without cache a few times before you actually see it, because the audio is so short

### DIFF
--- a/src/lib/features/playback-controls/ui/scrubbing-bar.svelte
+++ b/src/lib/features/playback-controls/ui/scrubbing-bar.svelte
@@ -1,11 +1,12 @@
 <script lang="ts">
   import { formatDuration } from '$lib/shared/ui';
+  import { renderTimeRanges } from '../lib/render-time-ranges';
 
   import { createEventDispatcher } from 'svelte';
 
   export let duration: number;
   export let position: number;
-  // export let buffered: TimeRanges;
+  export let buffered: TimeRanges | undefined;
 
   $: remaining = duration - position;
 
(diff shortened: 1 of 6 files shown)


src/lib/features/playback-controls/ui/scrubbing-bar.svelte
```diff
@@ -48,6 +55,11 @@
     --range: calc(var(--max) - var(--min));
     --ratio: calc((var(--value) - var(--min)) / var(--range));
     --sx: calc(0.5 * 18px + var(--ratio) * (100% - 18px));
+    --played-part: linear-gradient(#818cf8, #818cf8) 0 / var(--sx) 100% no-repeat;
+    --underlying-layer: #64748b;
```
alkaitagi, line 59: Question: wouldn't TailwindCSS be better for consistency?

alkaitagi (approved):

src/lib/features/playback-controls/ui/scrubbing-bar.svelte
illright: It certainly would, but in this particular case there's just no decent way to incorporate it. Not that I see, anyway.

alkaitagi: I see you use auto-generated code. Then since we won't have to manually edit these styles, I think it's fine to have plain CSS.

PR #97, opened 2022-03-05
Scrubbing fix

### LINKED ISSUE #96: Fix audio control scrubbing
Current when user drags the audio slider, apparently, the playback and cursor start fighting for controlling the position. This results in strong glitch.

A possible fix for this can be in implementing YouTube-like dragging behaviour: pause on start and play on end.

alkaitagi:
Closes #96.

### DIFF
--- a/src/lib/features/playback-controls/ui/scrubbing-bar.svelte
+++ b/src/lib/features/playback-controls/ui/scrubbing-bar.svelte
@@ -12,7 +12,7 @@
   $: remaining = duration - position;
 
   interface ScrubbingBarEvents {
-    scrub: { position: number };
+    scrub: { dragging: boolean; position: number };
   }
 
   const dispatch = createEventDispatcher<ScrubbingBarEvents>();
@@ -23,10 +23,12 @@
     type="range"
     min={0}
     max={duration}
-    step={0.01}
+    step={0.001}
     value={position}
     aria-valuetext={formatDuration(position)}
-    on:input={(e) => dispatch('scrub', { position: parseInt(e.currentTarget.value) })}
+    on:pointerup={() => dispatch('scrub', { dragging: false, position })}
+    on:input={(e) =>
+      dispatch('scrub', { dragging: true, position: parseInt(e.currentTarget.value) })}
     class="scrubbing-bar slider-progress w-full"
     style:--value={position}
     style:--min={0}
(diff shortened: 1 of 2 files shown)


src/lib/features/playback-controls/ui/scrubbing-bar.svelte
```diff
@@ -23,10 +23,12 @@
     type="range"
     min={0}
     max={duration}
-    step={0.01}
+    step={0.001}
```
illright, line 26: question (non-blocking): why did you increase the precision? I'm concerned for keyboard users because a single left/right arrow press moves the slider value one step, so they will have to press for very long if they want to scrub. 

suggestion (non-blocking): I actually think even `0.01` is too fine of a precision, and we'd better stick to `0.1`.

src/lib/features/playback-controls/ui/scrubbing-bar.svelte
```diff
@@ -23,10 +23,12 @@
...
     value={position}
     aria-valuetext={formatDuration(position)}
-    on:input={(e) => dispatch('scrub', { position: parseInt(e.currentTarget.value) })}
+    on:pointerup={() => dispatch('scrub', { dragging: false, position })}
```
illright, line 29: thought (non-blocking): looks good in that it probably works, but I can't shake the feeling of wanting something more elegant. My points of concern with this solution are the keyboard accessibility and the support of browsers regarding pointer events. The latter is kind of a nitpick since even IE supports pointer events, but the former one is a valid concern.

The crap thing is that I have no clue myself about how to implement this in a better way. Spent a good 10 minutes thinking of a way, and nothing came of it. For future reference, here's [a minimal reproduction of this issue](https://svelte.dev/repl/10118a4906d1479ab1ae9aae218ceeb2?version=3.46.4) in the Svelte REPL.

illright (approved):
I hope you tested it to work, but looks solid. That question about the precision though, but feel free to merge like that too

src/lib/features/playback-controls/ui/scrubbing-bar.svelte
alkaitagi: Because there is quite tangible snapping on drag end, but admittedly even after setting the step to a lower value it didn't go awaut.

PR #113, opened 2022-03-07
Add audio API specs

### LINKED ISSUE #102: Write test specs for `entities/audio`
**Acceptance criteria**:
- each segment in the slice should contain a `specs` directory that contains spec files for every module that is worth unit-testing
- the spec files contain a bunch of `it()`-s with a string description of the test and an empty function for the test body

alkaitagi:
Added 3 tests to cover recently updated audio API. Closes #102.


src/lib/entities/audio/specs/audio-instance.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { get } from 'svelte/store';
+import { src, currentTime, paused, move, pause, play, seek } from '..';
+
+describe('Audio playback API', () => {
+  const defaultSrc = 'http://commondatastorage.googleapis.com/codeskulptor-assets/Evillaugh.ogg';
+  const playDefault = () => play(defaultSrc);
+
+  it('load audio', () => {
```
illright, line 8: nitpick (non-blocking): the code would read slightly better like this:
```suggestion
  it('loads audio', () => {
```

aabounegm (commented):
I would recommend a few more tests, such as seeking to a negative time or beyond the audio duration

src/lib/entities/audio/specs/audio-instance.spec.ts
illright: I meant to do this for all `it()`-s :)

alkaitagi:
For some reason in tests audio loaing doesn't correctly set `duration`. I've rested it with logs in `npm run dev` and it works. Does the testing environment have no network connection?

illright:
The testing environment doesn't have a DOM at all, you're not mounting anything. I actually think it's for the best, and wherever you need internet or the DOM, I suggest you mock your way out of it

src/lib/entities/audio/specs/audio-instance.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+    pause();
+    play();
+    expect(get(paused)).toBeFalsy();
+    pause();
+    expect(get(paused)).toBeTruthy();
+  });
+
+  it('controls playback current time', async () => {
+    playDefault();
+    await new Promise((r) => setTimeout(r, 1000));
```
illright, line 27: issue: explicit hard-coded waits could lead to test flake and unnecessary slowdown of test execution

question: why do we need it? Can we mock stuff in order to remove the need for this wait?

illright (changes requested):

illright (approved):

alkaitagi:
@illright for some reason the actions passes but the `build-and-test` is in "expected" mode and I can't merge.

PR #115, opened 2022-03-08
Add empty specs

### LINKED ISSUE #105: Write test specs for `features/read-transcript`
**Acceptance criteria**:
- each segment in the slice should contain a `specs` directory that contains spec files for every module that is worth unit-testing
- the spec files contain a bunch of `it()`-s with a string description of the test and an empty function for the test body

### LINKED ISSUE #106: Write test specs for `widgets/now-playing`
**Acceptance criteria**:
- each segment in the slice should contain a `specs` directory that contains spec files for every module that is worth unit-testing
- the spec files contain a bunch of `it()`-s with a string description of the test and an empty function for the test body

khaledismaeel:
When completed, closes #105 and closes #106, though that'll be handled in a while.


src/lib/features/read-transcript/ui/specs/toggle-transcript.spec.ts
```diff
@@ -0,0 +1,5 @@
+export {};
+
+it.todo('displays the current transcript state to the user');
```
illright, line 3: issue (non-blocking): this will make you write two `render()` calls in the same test, which will make it grow excessively in terms of code size.

suggestion: split it into two tests — one for when the transcript is hidden and another for when it's shown

praise: wow, `it.todo()` is a thing? Didn't know that, great job discovering it

src/lib/features/read-transcript/ui/specs/toggle-transcript.spec.ts
```diff
@@ -0,0 +1,5 @@
...
+
+it.todo('dispatches the click event to the handler');
```
illright, line 5: nitpick (non-blocking): More like _to the parent_

src/lib/widgets/now-playing/ui/specs/now-playing.spec.ts
```diff
@@ -0,0 +1,5 @@
+export {};
```
illright, line 1: issue: you should also probably test that the `EpisodeDisplay` and `FullControls` are integrated correctly

suggestion: add a couple more tests that check that the episode title is present somewhere `inTheDocument` and the same for a button that says "Play"

illright (changes requested):
I'm not sure what you meant by the PR description, but this code you've written is _enough_ to close them both. However, in order for GitHub to understand that you want this PR to close them both, you need to write something like `Closes #X, closes #Y`. GitHub doesn't understand sentence semantics :)

src/lib/widgets/now-playing/ui/specs/now-playing.spec.ts
```diff
@@ -0,0 +1,9 @@
...
```
illright, line 1: issue: I think what's also necessary to test here is the presence of the speed toggle and the button to collapse the widget

illright (changes requested):

src/lib/widgets/now-playing/ui/specs/now-playing.spec.ts
khaledismaeel: The speed toggle is already included in playback controls, this will be tested in the fourth test. About the collapse button, yes, I will get to it.

illright (approved):

PR #118, opened 2022-03-08
Make the `notNull` function smarter

### LINKED ISSUE #98: Add a type guard to the `notNull` function

illright:
Closes #98

### DIFF
--- a/src/lib/shared/api/adapters.ts
+++ b/src/lib/shared/api/adapters.ts
@@ -35,6 +35,6 @@ export const transformPodcastRequest = (podcast: SBPodcast): Podcast | null => {
     coverUrl: coverUrlData.publicURL,
     title: podcast.title,
     author: podcast.author,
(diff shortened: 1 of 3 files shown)


aabounegm (changes requested):
Missing the assertion in `lib/shared/api/adapters.ts:38`:
https://github.com/aabounegm/cast/blob/4376fcf987819af54a56a68bb44dd0ea96f52a9c/src/lib/shared/api/adapters.ts#L38

illright:
@aabounegm fixed

aabounegm (approved):
I would have preferred using [`NonNullable<T>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#nonnullabletype) in the return type to make it more explicit (since `T` itself could originally contain `null`, but I guess it's fine for our current use case since type inference doesn't deduce `null` in the generic type as it is already in the type union in the argument type. In other words, the only case where it can cause problems is if we explicitly use `notNull<Podcast | null>(podcast);`

PR #121, opened 2022-03-10
Refactor FSD entities

### LINKED ISSUE #101: Move entity-related API code into `entities`
For example, the [`shared/api/podcasts.ts`](https://github.com/aabounegm/cast/blob/main/src/lib/shared/api/podcasts.ts) file belongs to `entities/podcast/api` in terms of FSD. The same for the episodes.

I also think that the adapters belong to entities, but the types should remain in shared. The reason why I want all of this is because `shared/api` currently looks overwhelming to me with a lot of files on the same level. It's harder to keep track of what's where

VanishMax:
Closes #101 
This PR replaces #117 it was easier to rewrite than to fix conflicting brances, restore the changes and fix them again

### DIFF
--- a/.github/workflows/unit-test-main.yml
+++ b/.github/workflows/unit-test-main.yml
@@ -13,6 +13,6 @@ jobs:
 
       - name: Run tests 🧪
         uses: ./.github/actions/unit-tests-action
-        env:
-          VITE_SUPABASE_URL: ${{ secrets.VITE_SUPABASE_URL }}
-          VITE_SUPABASE_KEY: ${{ secrets.VITE_SUPABASE_KEY }}
+        with:
+          supabaseUrl: ${{ secrets.VITE_SUPABASE_URL }}
+          supabaseKey: ${{ secrets.VITE_SUPABASE_KEY }}
--- a/.github/workflows/unit-test-main.yml
+++ b/.github/workflows/unit-test-main.yml
@@ -13,6 +13,6 @@ jobs:
 
       - name: Run tests 🧪
         uses: ./.github/actions/unit-tests-action
-        with:
-          supabaseUrl: ${{ secrets.VITE_SUPABASE_URL }}
-          supabaseKey: ${{ secrets.VITE_SUPABASE_KEY }}
+        env:
+          VITE_SUPABASE_URL: ${{ secrets.VITE_SUPABASE_URL }}
+          VITE_SUPABASE_KEY: ${{ secrets.VITE_SUPABASE_KEY }}
--- a/src/lib/shared/api/supabase.ts
+++ b/src/lib/shared/api/supabase.ts
@@ -1,3 +1,4 @@
+import 'isomorphic-fetch';
 import { createClient } from '@supabase/supabase-js';
 import constants from './credentials';
 
--- a/src/lib/shared/api/supabase.ts
+++ b/src/lib/shared/api/supabase.ts
@@ -1,4 +1,3 @@
-import 'isomorphic-fetch';
 import { createClient } from '@supabase/supabase-js';
 import constants from './credentials';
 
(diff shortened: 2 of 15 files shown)


VanishMax:
@illright take a look at how Jest is fixed:

1) I've added the `babel-vite` plugins, so the `import.meta.env` is changed to `process.env`
2) I've added `dotenv` package to parse `.env` files for Jest since Vite doesn't do it for us now
3) I've added `isomorphic-fetch`, so the `fetch` function is recognized in node environment

Locally, it works, but fails in CI because of the way we've added the variables. Take a look at the previous CI code:

```yaml
with:
  supabaseUrl: ${{ secrets.VITE_SUPABASE_URL }}
  supabaseKey: ${{ secrets.VITE_SUPABASE_KEY }}
```

As [GitHub documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepswith) states, `supabaseUrl` is transformed to `INPUT_SUPABASEURL` which is not what we want in the app. That is why I have changed the code to:

```yaml
env:
  VITE_SUPABASE_URL: ${{ secrets.VITE_SUPABASE_URL }}
  VITE_SUPABASE_KEY: ${{ secrets.VITE_SUPABASE_KEY }}
```

This cannot be tested before it gets to the `main` branch. Therefore, I believe my job is done here:)

src/lib/shared/api/supabase.ts
```diff
@@ -1,3 +1,4 @@
+import 'isomorphic-fetch';
```
illright, line 1: question: why is this necessary? It worked without it previously because SvelteKit already contains an isomorphic implementation of `fetch`

.github/workflows/unit-test-main.yml
```diff
@@ -13,6 +13,6 @@ jobs:
 
       - name: Run tests 🧪
         uses: ./.github/actions/unit-tests-action
-        with:
-          supabaseUrl: ${{ secrets.VITE_SUPABASE_URL }}
-          supabaseKey: ${{ secrets.VITE_SUPABASE_KEY }}
+        env:
```
illright, line 16: > As [GitHub documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepswith) states, supabaseUrl is transformed to INPUT_SUPABASEURL which is not what we want in the app.

thoughts: I don't think that's a point of concern because we explicitly give correct names to env vars [here](https://github.com/aabounegm/cast/blob/feature/fsd-entities-2/.github/actions/unit-tests-action/action.yml#L40-L41). It was working before, as far as I'm aware.

question: what exactly is the reason why you changed it in this way?

src/lib/shared/api/supabase.ts
VanishMax: Well, Jest gave the error `fetch is not defined`. So, I guess, it doesn't work

.github/workflows/unit-test-main.yml
VanishMax: Oh wait, I didn't even notice the `actions` folder. Ok, in this case I have no idea how to fix it.
You can try running this PR locally – it works. But not in the CI environment – `constants.SUPABASE_URL` is an empty string.

VanishMax: Ok, found. In the [`actions/unit-tests-action/action.yml`](https://github.com/aabounegm/cast/blob/feature/fsd-entities-2/.github/actions/unit-tests-action/action.yml#L40-L41) file, there were no Environmental variables next to the Unit testing job, only in the E2E tests

src/lib/shared/api/supabase.ts
illright: Aha, I see now why it worked previously. Yes, indeed it seems like Jest doesn't have `fetch`. However, I'm not a fan of including this into our actual codebase.

suggestions:
* move `isomorphic-fetch` to `devDependencies`
* remove the import of `isomorphic-fetch` from the application code
* add `'isomorphic-fetch/fetch-npm-node'` to the `setupFiles` in Jest

Just checked it, works equally well

illright (changes requested):
Let's get that `isomorphic-fetch` in order and we should be good to merge

VanishMax:
Done 👍

illright (approved):

PR #125, opened 2022-03-11
Login UI

### LINKED ISSUE #108: Implement the UI for logging in/out
Requires an artifact from #99, but most of the frontend code and the tests can be implemented independently.

**Acceptance criteria**:
- the _Your Library_ page should contain the button to log in or log out
- the design of the authentication element should look identical to the proposed design:
  image
- unit tests should be added for the code

Note: you don't have to implement the authentication flow itself, that is, no need to handle responses from Supabase, store the user data or anything. Create a widget that will display either of the states based on the value of a prop like `currentUser` or something.

VanishMax:
When implemented, closes #108 

Tried to follow the FSD as much as possible :)


src/lib/entities/user/index.ts
```diff
@@ -0,0 +1,2 @@
+export type { User } from './model/types';
+export { user } from './model/store';
```
illright, line 2: praise: hey, nice, we matched up in how we exposed our user store :)

src/lib/features/auth/specs/auth.spec.ts
```diff
@@ -0,0 +1,17 @@
+import { get } from 'svelte/store';
```
illright, line 1: hate to tell you this, but this won't be necessary :p. I implemented it in #126 (which you can review, btw)

src/lib/features/auth/specs/login-button.spec.ts
```diff
@@ -0,0 +1,11 @@
+import svelte from 'svelte-inline-compile';
+import { LoginButton } from '..';
+import { useLocalVars } from '$lib/shared/lib/jest-hacks';
+import { render } from '@testing-library/svelte';
+
+it('renders the button', async () => {
```
illright, line 6: issue: I think it's important for the user that the button is clickable and actually calls the authentication flow on press, which is not covered by the tests currently

src/lib/features/auth/specs/login-button.spec.ts
```diff
@@ -0,0 +1,11 @@
...
+  const loginButton = useLocalVars(svelte`<LoginButton />`, [LoginButton]);
```
illright, line 7: suggestion: no need for `svelte-inline-compile` here, you can just pass the `LoginButton` straight to the `render()` function

src/lib/pages/library/ui/library-page.svelte
```diff
@@ -2,12 +2,22 @@
   import { H1 } from '$lib/shared/ui';
   import { likesStore } from '$lib/features/like-episode';
   import { EpisodeCard } from '$lib/widgets/episode';
+  import { LoginContainer } from '$lib/widgets/user';
```
illright, line 5: issue: it sounds strange to me how a "login container" is inside a widget called "user":
* I don't really understand what is a login container
* I don't understand what could the "user" widget contain

suggestion: let's move the code that represents the user into `entities/user` and combine it with the buttons on the `pages` layer, I don't think we need widgets here

src/lib/widgets/user/index.ts
```diff
@@ -0,0 +1 @@
+export { default as LoginContainer } from './ui/login.svelte';
```
illright, line 1: suggestion (non-blocking): naming components exactly after their filename helps to navigate around the project, I suggest we call this file `login-container` (although, according to my previous comments, I don't think we need this file at all)

src/lib/widgets/user/ui/avatar.svelte
```diff
@@ -0,0 +1,7 @@
+<script lang="ts">
+  export let image: string;
+</script>
+
+<div class="aspect-square h-12 w-12 rounded-full overflow-hidden">
+  <img src={image} alt="user avatar" />
```
illright, line 6: issue (non-blocking): I'm not sure if this alt text is actually helpful. Alt text works best when it describes the content of an image, not its purpose, which is usually obvious from context

illright (changes requested):
I just submitted my PR on the authentication flow, it may be helpful for you to get it merged to avoid writing placeholders in place of authentication logic. So you're welcome to review #126 to get it merged faster :)

src/lib/features/auth/specs/login-button.spec.ts
VanishMax: I would really like if you provide an idea on how to actually do it

illright: You need to mock the `$lib/features/authenticate` module, trigger a click on a button with `@testing-library/user-event` and then check the mock to see if it has been called or not

VanishMax:
@illright I merged your PR into this one and connected UI with the logic. Everything's good except one thing: when I log in using GitHub, it redirects me back to the website, but the `user` writable store becomes filled only after reloading the page

src/lib/features/authenticate/ui/specs/sign-in-button.spec.ts
```diff
@@ -0,0 +1,7 @@
+import SignInButton from '../sign-in-button.svelte';
+import { render } from '@testing-library/svelte';
+
+it('renders the button', async () => {
+  const { getByRole } = render(SignInButton);
+  expect(getByRole('button', { name: 'Sign in with GitHub' })).toBeInTheDocument();
```
illright, line 6: question: so what about those mocks that we discussed above?

illright (commented):
issue (non-blocking): you didn't delete the `AudioFetch` file, you made it empty. I deleted it in my PR, so now this PR adds it back again

illright:
> but the user writable store becomes filled only after reloading the page

That's weird, that should not happen. Oh well, let's keep it this way for now and address this problem in #127

src/lib/features/authenticate/ui/specs/sign-out-button.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+  signOut: jest.fn(),
+}));
+
+import { signOut } from '../../model/sign-out';
+
+it('renders and clicks the sign out button', async () => {
+  const user = userEvent.setup();
+  const { getByRole } = render(SignOutButton);
+  await user.click(getByRole('button', { name: 'Sign out' }));
+  expect(signOut).toHaveBeenCalledTimes(1);
```
illright, line 15: nitpick (non-blocking): I think the number of times this function has been called is not that important in this case, there's a shorter `.toHaveBeenCalled()` assertion. But this works too, and asserts even more things

src/lib/features/authenticate/ui/specs/sign-out-button.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+import { render } from '@testing-library/svelte';
+import userEvent from '@testing-library/user-event';
+
+jest.mock('../../model/sign-out', () => ({
+  signOut: jest.fn(),
+}));
+
+import { signOut } from '../../model/sign-out';
+
+it('renders and clicks the sign out button', async () => {
```
illright, line 11: nitpick (non-blocking): I like to read these `it`-s as being about the filename. So saying that a "sign-out-button" renders and clicks the sign out button feels strange to me. I'd phrase it as a "sign-out-button" calls the `signOut` function on click

src/lib/features/authenticate/ui/specs/sign-out-button.spec.ts
```diff
@@ -0,0 +1,16 @@
+import SignOutButton from '../sign-out-button.svelte';
+import { render } from '@testing-library/svelte';
+import userEvent from '@testing-library/user-event';
+
+jest.mock('../../model/sign-out', () => ({
+  signOut: jest.fn(),
+}));
+
+import { signOut } from '../../model/sign-out';
```
illright, line 9: thoughts (non-blocking): by the way, Jest hoists the `jest.mock` calls so you don't actually have to import after the mock. But I see why you'd want to do this

illright (approved):
praise: the tests are good! Clean and accessible

Not sure what's wrong with the Cloudflare preview, it showed me the button once, I logged in, then everything related to `UserContainer` has just disappeared.

illright:
Oh, I understood what's wrong with the Cloudflare demo. I logged in from the preview deployment, but it redirected me to the stable deployment. That's a shame, but whatever

PR #126, opened 2022-03-11
Authentication flow

### LINKED ISSUE #109: Implement the authentication flow
Blocked by #99 

Acceptance criteria:
- the user should be directed to a dedicated endpoint from the backend upon a sign in request (no need to actually implement/style the button)
- the token should be accepted from Supabase and stored in a persistent manner
- the token should automatically be renewed when it's nearing expiration
- the API code should support token passing

illright:
Closes #109. Supabase has made it too easy, too easy.

### DIFF
--- a/src/lib/entities/user/model/store.ts
+++ b/src/lib/entities/user/model/store.ts
@@ -1,0 +1,5 @@
+import { writable } from 'svelte/store';
+import type { User } from '@supabase/supabase-js';
+import { supabaseClient } from '$lib/shared/api';
+
+export const user = writable<User | null>(supabaseClient.auth.user());
--- a/src/lib/features/authenticate/model/track-auth-status.ts
+++ b/src/lib/features/authenticate/model/track-auth-status.ts
@@ -1,0 +1,15 @@
+import noop from 'lodash.noop';
+import { user } from '$lib/entities/user';
+import { supabaseClient } from '$lib/shared/api';
+
+export function trackAuthStatus() {
+  const { data, error } = supabaseClient.auth.onAuthStateChange((_event, session) => {
+    user.set(session?.user ?? null);
+  });
+
+  if (error !== null) {
+    throw error;
+  }
+
+  return data?.unsubscribe ?? noop;
+}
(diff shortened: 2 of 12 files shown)


src/lib/entities/user/model/store.ts
```diff
@@ -0,0 +1,5 @@
+import { writable } from 'svelte/store';
+import type { User } from '@supabase/supabase-js';
+import { supabaseClient } from '$lib/shared/api';
+
+export const user = writable<User | null>(supabaseClient.auth.user());
```
VanishMax, line 5: Question: in SvelteKit, does this line mean that authentication will happen on the server or after the page is loaded on client? Or, in other words, will it cause the FOUC?

src/lib/features/authenticate/model/track-auth-status.ts
```diff
@@ -0,0 +1,15 @@
...
+export function trackAuthStatus() {
+  const { data, error } = supabaseClient.auth.onAuthStateChange((_event, session) => {
+    user.set(session?.user ?? null);
+  });
+
+  if (error !== null) {
+    throw error;
+  }
+
+  return data?.unsubscribe ?? noop;
```
VanishMax, line 14: Question: I see `noop` is used only as the funtion returned from the `trackAuthStatus` function, but the return data is never actually used in the code. So, do we need to have `noop` intalled as a dependency if it isn't used?

VanishMax (commented):
Praise: the tests are fairly good, I always find something to learn in them

src/lib/entities/user/model/store.ts
illright: It's not gonna cause FOUC because it has nothing to do with CSS, but it will render as logged out first and then catch up immediately. I think we can do better, but that's probably going to require some research about how CloudFlare SSR works. I think it's worth another issue

src/lib/features/authenticate/model/track-auth-status.ts
illright: It is used — the function that you return from an `onMount` callback will be called on unmounting to clean up. And this dependency is so tiny I don't think it's ever going to be an issue.

It's just that we have ESLint set up to forbid empty functions (which is good, if I think about it)

src/lib/entities/user/model/store.ts
VanishMax: Ok, I understand now. I guess the term [Cumulative Layout Shift (CLS)](https://web.dev/cls/) describes the situation better

VanishMax (approved):

PR #128, opened 2022-03-12
Snackbar feature

### LINKED ISSUE #107: Implement a snackbar notification system
Maybe [`renderless-svelte`](https://github.com/stephane-vanraes/renderless-svelte) will come in handy here.

**Acceptance criteria**:
- the `features` layer should have access to a function that initiates a snackbar display
- the snackbar should stay displayed for 4000ms and then disappear automatically
- the snackbar should be able to display an arbitrary text message and support a button with a customizable text to call a passed callback on click
- unit tests are provided
- the snackbar looks identical to the proposed design:
  image

VanishMax:
Snackbar implementation with the queue, closes #107 

It allows pushing notifications to user by the `snackbar()` function. If the `snackbar` is called again with the same title, it will be prolonged. If `snackbar` is called with another title, it will be added to the end of the queue, wait till the first element is done, and show up on the screen.

This time I finally succeeded with tests 💪


src/lib/features/snackbar/model/store.ts
```diff
@@ -0,0 +1,47 @@
+import { writable, get } from 'svelte/store';
+import type { SnackbarOptions } from './types';
+
+/**
+ * The queue of different snackbars. It works like this:
+ *
+ * • First comes served – the first snackbar stays on the screen for 4 seconds
```
illright, line 7: nitpick (non-blocking): These bullet-point characters are neat, but JSDoc is Markdown, so they won't be rendered into proper HTML lists. In this way, it's better to use dashes `-`

src/lib/features/snackbar/model/store.ts
```diff
@@ -0,0 +1,47 @@
...
+ * didn't pass but the call for this type of snackbar fired again, timeout resets
+ * to 4 seconds
+ * • Pressing on the Snackbar's action removes it from the queue
+ */
+export const snackbarQueue = writable<SnackbarOptions[]>([]);
+
+const SNACKBAR_TIMEOUT = 4000;
+let timeout: ReturnType<typeof setTimeout>;
+
+// TODO: экспортировать фильтр и сделать обертку для каждого колбека
```
illright, line 19: question: what's this?)

src/lib/features/snackbar/ui/snackbar-queue.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+  import { snackbarQueue } from '../model/store';
+  import Snackbar from './snackbar.svelte';
+
+  $: activeSnackbar = $snackbarQueue[0] || null;
+</script>
+
+{#if activeSnackbar}
+  <Snackbar snackbar={activeSnackbar} />
```
illright, line 9: nitpick (non-blocking): a lot of repetition in this line, I'd suggest naming the prop `options` like you named your type

src/lib/features/snackbar/index.ts
```diff
@@ -0,0 +1,3 @@
+export type { SnackbarOptions } from './model/types';
```
illright, line 1: [about this whole slice] 

issue: I don't think `snackbar` is a feature of our application. Snackbars by itself don't really add any business logic to us.

suggestion: move this slice into `shared/ui`

src/lib/features/snackbar/ui/snackbar.svelte
```diff
@@ -0,0 +1,23 @@
...
+  export let snackbar: SnackbarOptions;
+
+  const callback = () => {
+    snackbar.callback();
+    filterQueueByText(snackbar.text);
+  };
+</script>
+
+<div
+  class="fixed bottom-16 mb-2 left-2 right-2 border border-white border-2 rounded-lg p-4 flex flex-col bg-background-slightly-higher"
```
illright, line 14: nitpick (non-blocking): `left-2 right-2` is not going to position it 8px from the left and right sides, it's gonna be one or the other. In this case it doesn't matter, I think

src/lib/features/snackbar/model/store.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const filterQueueByText = (text: string) => {
+  snackbarQueue.update((queue) => {
+    return queue.filter((snack) => snack.text !== text);
```
illright, line 23: issue (non-blocking): this is not as efficient as it could be, since this function will only ever be used to remove the first snackbar in the queue (from what it seems). But since this queue is unlikely to contain more than 3 elements at a time, it's not a big deal.

src/lib/features/snackbar/model/store.ts
```diff
@@ -0,0 +1,47 @@
+import { writable, get } from 'svelte/store';
+import type { SnackbarOptions } from './types';
+
+/**
+ * The queue of different snackbars. It works like this:
```
illright, line 5: praise: thanks for documenting this, really helped me to understand this file

src/lib/features/snackbar/specs/snackbar.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+  expect(queryByText(secondSnackbar.text)).not.toBeInTheDocument();
+});
+
+it('removes snackbar from the queue after pressing the button', async () => {
+  const user = userEvent.setup();
+  snackbar(firstSnackbar);
+  const { getByRole, queryByText } = render(SnackbarQueue);
+
+  expect(queryByText(firstSnackbar.text)).toBeInTheDocument();
+  await user.click(getByRole('button', { name: firstSnackbar.buttonText }));
```
illright, line 37: suggestion: I think it's a good idea to also assert here that the callback was called

src/lib/features/snackbar/specs/snackbar.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+const secondSnackbar: SnackbarOptions = {
+  text: 'Example',
+  buttonText: 'Close',
+  callback: jest.fn(),
+};
+
+it('adds a snackbar to the queue and displays it', () => {
+  snackbar(firstSnackbar);
+  const { getByText } = render(SnackbarQueue);
+  expect(getByText(firstSnackbar.text)).toBeInTheDocument();
```
illright, line 20: issue: usually, things that rely on time are particularly fussy to test, the tests for them tend to be flaky. Jest has a way to remedy that by letting you control time. See [this page](https://jestjs.io/docs/timer-mocks) for reference.

suggestion: add time control to avoid test flake

illright (changes requested):
praise: looks nice, I checked it out on the preview

praise: the tests are good too, great job harnessing the Testing Library)

suggestion: I think we could add to the test suite though, for instance, test that the snackbars actually disappear by themselves. I think it's pretty important from a customer POV

src/lib/features/snackbar/model/store.ts
aabounegm: Also, I don't understand when people write `(x) => { return ...; }` instead of simply `(x) => ...`, which is what arrow functions are mainly for

illright: @aabounegm that sounds kind of harsh

src/lib/features/snackbar/model/store.ts
VanishMax: oups, it was a message for myself 😄

src/lib/features/snackbar/index.ts
VanishMax: But why to `shared/ui`? If would feel strange to have `shared/ui/snackbar/model/store.ts` file

illright: True, you can move the store outside of the UI segment and into the model in shared. It is just not a feature, in my opinion

src/lib/features/snackbar/specs/snackbar.spec.ts
VanishMax: Good idea, implemented this 👍

illright: But why did you only do it for one test? It's best if all tests will be time-independent

src/lib/shared/ui/snackbar/store.ts
```diff
@@ -0,0 +1,45 @@
...
+  timeout = setTimeout(() => {
+    filterQueueByText(options.text);
+    const queue = get(snackbarQueue);
+    if (queue.length) {
+      setSnackbarTime(queue[0]);
+    }
+  }, SNACKBAR_TIMEOUT);
+};
+
+export const snackbar = (options: SnackbarOptions) => {
```
aabounegm, line 36: It is better not to give functions nouns as names. I'd prefer some imperative form, such as `showSnackbar`

src/lib/shared/ui/snackbar/store.ts
```diff
@@ -0,0 +1,45 @@
...
+ * - First comes served – the first snackbar stays on the screen for 4 seconds
+ * - Next snackbar waits in the queue for its turn
+ * - Snackbar timing can be prolonged – if the time of the first snackbar in the queue
+ * didn't pass but the call for this type of snackbar fired again, timeout resets
+ * to 4 seconds
+ * - Pressing on the Snackbar's action removes it from the queue
+ */
+export const snackbarQueue = writable<SnackbarOptions[]>([]);
+
+const SNACKBAR_TIMEOUT = 4000;
```
aabounegm, line 16: Why isn't this configurable for each snackbar? We could have it as part of the options but with a default value

src/lib/shared/ui/snackbar/types.d.ts
```diff
@@ -0,0 +1,5 @@
+export interface SnackbarOptions {
+  text: string;
+  buttonText: string;
```
aabounegm, line 3: How about `actionText` as a name? it makes the purpose of the button clearer

src/lib/shared/ui/snackbar/types.d.ts
```diff
@@ -0,0 +1,5 @@
...
+  callback: () => void;
```
aabounegm, line 4: Also just `callback` is too generic. It only gets called when the close button is clicked, so how about `closeCallback` or so?

src/lib/shared/ui/snackbar/snackbar-queue.svelte
```diff
@@ -0,0 +1,10 @@
+<script lang="ts">
+  import { snackbarQueue } from './store';
+  import Snackbar from './snackbar.svelte';
+
+  $: activeSnackbar = $snackbarQueue[0] || null;
+</script>
+
+{#if activeSnackbar}
+  <Snackbar options={activeSnackbar} />
```
aabounegm, line 9: suggestion (non-blocking): I'd really love if there was some kind of transition here

src/lib/shared/ui/snackbar/snackbar.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { render } from '@testing-library/svelte';
```
aabounegm, line 1: Shouldn't this file be under s folder called `specs`?

src/lib/shared/ui/snackbar/store.ts
```diff
@@ -0,0 +1,45 @@
...
+let timeout: ReturnType<typeof setTimeout>;
+
+export const filterQueueByText = (text: string) => {
```
aabounegm, line 19: So if a snackbar is queued multiple times, with different snackbars in between, all the others will be removed?
Note: you can use `array.shift()` to remove the first element of an array

src/lib/shared/ui/snackbar/store.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+export const snackbar = (options: SnackbarOptions) => {
+  snackbarQueue.update((queue) => {
+    if (!queue.length || queue[0]?.text === options.text) {
+      setSnackbarTime(options);
+    }
+
+    if (queue[0]?.text === options.text) return queue;
+    return [...queue, options];
```
aabounegm, line 43: Consider using `queue.push(options)` instead of copying the entire array

src/lib/pages/podcast-gallery/ui/index.svelte
```diff
@@ -26,3 +42,24 @@
     <PodcastList podcasts={$podcasts} />
   {/if}
 </div>
+
+<!--
+TODO: this is only to show the snackbar in action.
+  Remove after connecting the snackbars with other components
+-->
+<div class="flex px-4 py-2 gap-2">
```
aabounegm, line 50: Should this really be merged into `main`? I don't think it would look nice in the presentation in 2 days

src/lib/shared/ui/snackbar/store.ts
illright: I don't think there's a need for this, we're not building a UI library

src/lib/shared/ui/snackbar/snackbar-queue.svelte
illright: We'll have an entire sprint dedicated to animation and eye candy, I think it's best if we wait until then

src/lib/shared/ui/snackbar/store.ts
VanishMax: It cannot be queued multiple times due to the prolongation mechanism

src/lib/shared/ui/snackbar/snackbar.spec.ts
VanishMax: Why should it be there? FSD doesn't require the `specs` folder

VanishMax:
Most of the comments are the nitpicks about the function names. I hear you and consider these namings for the future PRs, but can we merge this now? 🙈

src/lib/shared/ui/snackbar/snackbar.spec.ts
illright: In other parts of this project we put segment tests into the `specs` folder inside each segment. Just for consistency with the rest of the project

illright:
@VanishMax for me there's still one blocking issue about tests and their time management. See the unresolved blocking comment above

illright (approved):

PR #143, opened 2022-03-18
PWA

### LINKED ISSUE #134: Make our application into a PWA
**Acceptance criteria**:
- the application should ship with a manifest for PWA (and thus, be installable)
- the app icons should be exported from Figma in the necessary resolutions
- when the app is offline, the Podcast Gallery page should display an empty state below the "Recently Listened" podcast shelf that says that the user is not connected to the internet (design the empty state yourself, I believe in you)
- the service worker should be configured to intercept requests for pages to serve them from cache

For inspiration, see one of my other SvelteKit apps with offline capabilities: https://github.com/illright/draw-ill-help/blob/main/src/service-worker.ts

There's also a set of packages by Google called Workbox to abstract away boring stuff about the service worker, feel free to research.

aabounegm:
Closes #134


src/service-worker.ts
```diff
@@ -0,0 +1,79 @@
...
+
+worker.addEventListener('activate', (event) => {
+  event.waitUntil(
+    caches.keys().then(async (keys) => {
+      // delete old caches
+      for (const key of keys) {
+        if (key !== cacheName) await caches.delete(key);
+      }
+
+      worker.clients.claim();
```
illright, line 31: issue (non-blocking?): I'm not sure but I think this code will not wait until the previous service workers are deactivated to activate the current one. This might be a problem, or it might not be. Not sure

src/service-worker.ts
```diff
@@ -0,0 +1,79 @@
...
+      .then(() => worker.skipWaiting())
+  );
+});
+
+worker.addEventListener('activate', (event) => {
+  event.waitUntil(
+    caches.keys().then(async (keys) => {
+      // delete old caches
+      for (const key of keys) {
+        if (key !== cacheName) await caches.delete(key);
```
illright, line 28: issue (non-blocking): it's deleting those caches sequentially while it could delete them in parallel :)

src/app.html
```diff
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf-8" />
     <meta name="description" content="Svelte demo app" />
+    <link rel="manifest" href="/manifest.json" />
     <link rel="icon" href="%svelte.assets%/favicon.png" />
```
illright, line 7: suggestion (non-blocking): maybe we could also change this favicon in this PR?) I've made a rounded version without shadows in Figma, you can grab it there if you want to

illright (approved):
praise: looks solid

thoughts: have you checked out the code that I've written [for service workers in `draw-ill-help`](https://github.com/illright/draw-ill-help/blob/main/src/service-worker.ts)? I like it a bit more because it gets to take full advantage of the `async`-`await` interface and also has very descriptive names for things that the service worker does. I don't necessarily mean you should adopt that style but let me know: which one looks better to you and why?

src/service-worker.ts
aabounegm: JavaScript (including workers) is single-threaded, so I wouldn't really call it _parallel_, but yeah, we don't know how the native functions are implemented under the hood, so I guess it wouldn't harm.

src/service-worker.ts
aabounegm: `claim` does not _activate_ the service worker. It is already activated, given that this code is inside the `activate` event handler. It is used exactly like that in the [`clients.claim` example on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Clients/claim#example).

aabounegm:
I checked your code only after having already added mine, and I do like that it is well documented and easy to understand. The one little thing I didn't like is that each function is specific to exactly one event and cannot be used otherwise, which doesn't really qualify it to be a separate function in my book).
The pros are probably more than the con(s) in this case, but I'm not sure it's worth the effort. Feel free to modify it yourself in another PR, but keep in mind that yours follows _cache-first_ strategy, while mine follows _online-first_.

PR #144, opened 2022-03-19
Playback controls rework

### LINKED ISSUE #138: Add the like button and speed control to the playback controls on the Now Playing page
Currently, the playback controls are coming from the `FullControls` feature component, but likes are another feature. We should break up the `FullControls` component into `ScrubbingBar` (exists) and `PlaybackControlButtons` (should be added). The tests that were written for the full controls should be split between the resulting two components.

**Acceptance criteria**:
- the Now Playing widget contains the button to like the currently playing episode
- the Now Playing widget contains the button to cycle through 0.5x, 1x, 1.5x, 2.0x playback speeds
- the controls should follow the proposed design
- Feature-Sliced Design principles should be respected

alkaitagi:
Closes #138.

Things I need advice with:
* Non-hacky way to control like-button size via classes.
* Was it meant in the issue description that `full-controls.svelte` needs to be gone completely?


src/lib/features/like-episode/ui/like-button.svelte
```diff
@@ -8,14 +8,17 @@
...
+  export let size = 6;
+  export let iconSize = 4;
+
   $: liked = $likesStore.has(episode.id);
 </script>
 
 <IconButton
   name={liked ? 'Remove the like' : 'Like this episode'}
-  class={clsx('w-6 h-6', liked && 'text-pink-500')}
+  class={clsx(`w-${size} h-${size}`, liked && 'text-pink-500')}
```
illright, line 19: suggestion: How about we just expose a `class` prop that will go after the default sizing, overriding it if necessary?

illright: And `iconClass` as well, that will just go straight to the icon

src/lib/widgets/bottom-bar/ui/mini-player.svelte
```diff
@@ -1,12 +1,8 @@
 <script lang="ts">
-  import IconReplay10 from '~icons/ic/round-replay-10';
-  import IconForward30 from '~icons/ic/round-forward-30';
-  import { PlaybackButton, toggleGlobalPlayback } from '$lib/features/playback-controls';
-  import { IconButton } from '$lib/shared/ui';
-  import { move, paused } from '$lib/entities/audio';
   import { currentlyPlayingEpisode } from '$lib/entities/episode';
   import { getPodcastByID } from '$lib/entities/podcast';
   import { NowPlaying } from '$lib/widgets/now-playing';
+  import PlaybackControlButtons from '$lib/features/playback-controls/ui/playback-control-buttons.svelte';
```
illright, line 5: issue: FSD forbids going inside of slices with imports

src/lib/widgets/bottom-bar/ui/mini-player.svelte
```diff
@@ -25,33 +21,5 @@
...
-      icon={IconForward30}
-      class="w-9 h-9"
-      iconClass="w-7 h-7"
-      on:click={(e) => {
-        e.stopPropagation();
-        move(30);
-      }}
-    />
-  </div>
+  <PlaybackControlButtons />
```
illright, line 24: thoughts: I wonder if this dependency is justified. I think that this duplication here does not cause too much trouble, but we retain the ability of customizing the appearance of the Now Playing page without breaking the mini player. Namely, the `prominent` prop makes sense for the playback button, but makes less sense for the full controls.

suggestion: keep this code as it was

src/lib/features/playback-controls/ui/playback-control-buttons.svelte
```diff
@@ -0,0 +1,36 @@
...
+    name="Skip forward 30 seconds"
+    icon={IconForward30}
+    class="w-9 h-9"
+    iconClass="w-7 h-7"
+    on:click={(e) => {
+      e.stopPropagation();
+      move(30);
+    }}
+  />
+  <slot />
```
illright, line 35: issue (non-blocking): this inconsistency in slot naming is not very pretty, in my opinion. I'd prefer `before` and `after`, maybe.

suggestion: better yet, I don't think this component should have slots. The composition can be done just by placing the respective components side by side. I understand that this might break the Flexbox arrangement, but that's okay.

src/lib/features/playback-controls/ui/full-controls.svelte
```diff
@@ -32,21 +36,20 @@
...
-      iconClass="w-7 h-7"
-      on:click={() => move(30)}
-    />
-  </div>
+  <PlaybackControlButtons {prominent}>
+    <button
+      slot="pre"
+      name="Toggle playback speed"
+      class="flex items-center justify-center w-10 h-10 font-semibold text-sm rounded-full hover:bg-slate-600 active:bg-slate-500"
+      on:click={(e) => {
```
illright, line 44: suggestion (non-blocking): when adding event handlers to HTML nodes, we can make use of Svelte event modifiers like so:
```suggestion
      on:click|stopPropagation={() => {
```

Then we can remove the line from the handler.

question: why do we even need to stop propagation here? It was useful in the mini-player, but I don't think we need it here

src/lib/features/playback-controls/ui/full-controls.svelte
```diff
@@ -1,24 +1,28 @@
...
+  import PlaybackControlButtons from './playback-control-buttons.svelte';
+  import LikeButton from '$lib/features/like-episode/ui/like-button.svelte';
+  import type { Episode } from '$lib/shared/api/types';
 
   let _class = '';
   export { _class as class };
+  export let prominent = false;
+  export let episode: Episode | undefined;
+
+  function nextSpeed() {
```
illright, line 22: nitpick (non-blocking): I'd prefer an "action" name like `cyclePlaybackSpeed`

src/lib/features/playback-controls/ui/full-controls.svelte
```diff
@@ -1,24 +1,28 @@
...
+    $speed = ($speed + 0.5) % 2.5;
```
illright, line 23: issue (non-blocking): I understand that this probably works, but it updates the store twice and just generally looks very complex

suggestion (non-blocking): maybe we could have an array of options and use that for cycling?

src/lib/features/playback-controls/ui/full-controls.svelte
```diff
@@ -1,24 +1,28 @@
...
     play,
+    playbackRate as speed,
   } from '$lib/entities/audio';
-  import { IconButton } from '$lib/shared/ui/';
-  import PlaybackButton from './playback-button.svelte';
-  import { toggleGlobalPlayback } from '../model/toggle-global-playback';
   import ScrubbingBar from './scrubbing-bar.svelte';
   import clsx from 'clsx';
+  import PlaybackControlButtons from './playback-control-buttons.svelte';
+  import LikeButton from '$lib/features/like-episode/ui/like-button.svelte';
```
illright, line 14: issue: FSD public API constraint here too

illright (changes requested):
To your question: I intended for the `FullControls` to disappear completely, but this was just my vision, you don't have to do that. But I still think we don't need it, though.

src/lib/widgets/now-playing/ui/now-playing.svelte
```diff
@@ -30,7 +46,27 @@
...
+      buffered={$buffered}
+      on:scrub={(e) => {
+        if (e.detail.dragging) pause();
+        else play();
+        seek(e.detail.position);
+      }}
+    />
+    <div class="w-full mt-5 flex justify-around">
+      <button
+        name="Toggle playback speed"
```
illright, line 61: issue (non-blocking): the action is not really "toggle", more like "cycle", as in the function name. I'm thinking, maybe we could even write the next speed in the button name? Only do it if you feel like doing it, it's a lot of code for little benefit

src/lib/features/playback-controls/ui/playback-controls.svelte
```diff
@@ -0,0 +1,34 @@
+<script>
+  import { paused, move } from '$lib/entities/audio';
+  import { toggleGlobalPlayback } from '..';
+  import { IconButton } from '$lib/shared/ui/';
+  import PlaybackButton from './playback-button.svelte';
+  import IconReplay10 from '~icons/ic/round-replay-10';
+  import IconForward30 from '~icons/ic/round-forward-30';
+
+  export let prominent = false;
```
illright, line 9: question: is this component used anywhere without this prop? I think we reverted the mini-player, so this prop isn't really necessary anymore

illright (approved):
I've left a few comments for changes, but they aren't blocking, feel free to ignore. However, the tests aren't passing for some reason, idk what's up with that. Maybe it has something to do with Supabase keys, maybe it doesn't. Could you investigate, please?

alkaitagi:
From the old `full-controls.spec.ts` file, the first 3 tests were moved into `playback-controls.spec.ts`, and the next 2 were removed because their seeking checks are already covered in `scrubbing-bar.spec.ts`.

PR #145, opened 2022-03-19
Try setting up the hero animation and fail

### LINKED ISSUE #133: Implement a hero animation for the podcast tile
A hero animation is when you click on a podcast tile that contains an image, and that image grows in size and moves to the place where that image will be located on the next page. Basically, it's a fancy smooth page transition animation

**Acceptance criteria**: 
- clicking a podcast tile should trigger the page transition (already does)
- the cover art of the podcast should smoothly scale and translate into the place of the cover art in the Podcast Episodes page

Note: no need for such a transition backwards (i.e. from Podcast Episodes to Podcast Gallery), but if it's easy to implement then by all means go for it

VanishMax:
It will never close #133 

Well, It was supposed to be easy in Svelte... But not in SvelteKit. The problem is that for the `crossfade` function (the only solution suitable for hero images) all contents of `_layout.svelte` is recreated next to the previous layout. You can try clonning this PR, tap on a podcast link and notice animation working. But check devtools: the are two `.layout`. The more you walk through the app, the more layouts you'll see. Somehow, after animation finished, it doesn't get removed. Believe me, I've tried a lot. 

Referencing everything I tried:
* [Dev.to article](https://dev.to/giorgosk/smooth-page-transitions-in-layout-svelte-with-sveltekit-or-sapper-4mm1). Inside of it, the author states _"Perhaps if anyone know how to add crossfade to it please do let me know in the comments"_.
* [Svelte REPL, showing `crossfade` function usage](https://svelte.dev/repl/0ad58a0d830f4001b91409e40164aa24?version=3.44.1)
* [StackOverflow answer](https://stackoverflow.com/questions/69896477/svelte-page-transition-duplicating-entire-site-layout-svelte) about the `crossfade` problem that no one knows how to fix without introducing new bugs

### DIFF
--- a/src/lib/entities/podcast/ui/podcast-display.svelte
+++ b/src/lib/entities/podcast/ui/podcast-display.svelte
@@ -1,12 +1,20 @@
 <script lang="ts">
   import { H1, s } from '$lib/shared/ui';
   import type { Podcast } from '$lib/shared/api';
+  import { send, receive } from '$lib/shared/ui/animation';
 
   export let podcast: Podcast;
 </script>
 
 <div class="flex flex-row gap-4 p-2">
-  <img class="w-24 h-24 rounded-xl" src={podcast.coverUrl} alt="" />
+  <img
+    class="w-24 h-24 rounded-xl"
+    src={podcast.coverUrl}
+    alt=""
+    in:receive={{ key: podcast.id }}
+    out:send={{ key: podcast.id }}
+  />
+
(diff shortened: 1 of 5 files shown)


illright:
Khaled and I took a look at this, we can't reproduce what you're talking about. In both Chrome and Firefox, local and Cloudflare, there is always only one instance of `.layout` in the DOM

PR #146, opened 2022-03-19
Fix the leak of service keys

khaledismaeel:


aabounegm (changes requested):
For consistency's sake, can you please rename them in the GitHub Action as well?

illright:
@aabounegm done

aabounegm (approved):

PR #149, opened 2022-03-23
Fix env var names

khaledismaeel:
Apparently, we broke `supabaseClient` by renaming the env vars in #146; now running the server locally `supabaseClient` is `undefined`. This should fix it.


illright:
But why though? #146 is already merged

khaledismaeel:
You're right; I just didn't pull main locally. Closing.

PR #150, opened 2022-03-24
Draggability of Now Playing widget

### LINKED ISSUE #132: Implement the draggability of the bottom bar
**Acceptance criteria**:
- in the Now Playing widget, a drag indicator should be present at the top (as per the design)
- dragging the Now Playing widget down on that indicator should slide it down
- past a certain threshold, it should slide down to the end automatically, before that threshold it should spring back into place (determine the threshold yourself)
- dragging should cause the Now Playing UI to fade out and the bottom bar UI to fade in, based on the current drag position
- the button to collapse the bottom bar should also run the same animation, but the position of the bottom bar will be changed automatically
- the Now Playing widget should not extend to the top of the screen, but only as far as it needs to go. Behind it should be an overlay over the page that is currently shown (see the design)

VanishMax:
Closes #132


VanishMax:
Ah, forgot about tests... If you are reviewing, please check the logic and code first

alkaitagi (approved):

src/lib/widgets/now-playing/ui/episode-display.svelte
```diff
@@ -4,7 +4,10 @@
   export let podcast: Podcast;
 </script>
 
-<img class="flex rounded-2xl" src={podcast.coverUrl} alt="" />
+<div
```
illright, line 7: question: why is this necessary?

VanishMax: There was a problem of image always taking the full width of the container and scaling vertically to enormous sizes. I wanted to try solving it with `img` tag but failed, so the `background` came up to place perfectly.

VanishMax:
@alkaitagi @illright You might want to check this PR again – I've added tests and refactored the code, since the were many changes from the `main` branch

src/lib/widgets/now-playing/lib/detect-swipe.ts
```diff
@@ -0,0 +1,66 @@
+const PIXEL_THRESHOLD = 60;
+
+type SwipeCallbacks = Partial<Record<'left' | 'right' | 'up' | 'down', () => void>>;
+
+export const detectSwipe = (callbacks: SwipeCallbacks) => {
```
illright, line 5: suggestion (non-blocking): would be even cooler to make this a Svelte action, something like this:
```svelte
<div use:swipe={{ down: () => dispatch('minimize') }} />
```

src/lib/widgets/now-playing/ui/swipe-down-bar.svelte
```diff
@@ -0,0 +1,21 @@
...
+  const dispatch = createEventDispatcher<{ minimize: void }>();
+
+  const { handlePointerMove, handlePointerStart, handlePointerUp } = detectSwipe({
+    down: () => {
+      dispatch('minimize');
+    },
+  });
+</script>
+
+<button
```
illright, line 14: question: why is it a button? Looks like this will produce very confusing screen reader semantics

src/lib/widgets/now-playing/ui/now-playing.svelte
```diff
@@ -54,6 +55,7 @@
           />
         {/if}
         <div class="flex flex-col items-center" class:invisible={transcriptShown}>
+          <SwipeDownBar on:minimize={() => nowPlayingActive.set(false)} />
```
illright, line 58: issue: I think we still want to show the drag indicator even when the transcript is shown

src/lib/widgets/now-playing/specs/swipe-down-bar.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+   * This logic is a mix of Testing Library `createEvent` function and an article
+   * https://www.freecodecamp.org/news/how-to-write-better-tests-for-drag-and-drop-operations-in-the-browser-f9a131f0b281/.
+   *
+   * Basically, it is a normal practice to test drag-and-drop (or pointerdown and pointermove) features
+   * by firings proper events with corresponding data (in this case, clientY property), but Testing
+   * Library's `createEvent` cannot assign the properties to Event entity. That's why the `Object.assign` is used.
+   */
+  fireEvent(
+    screen.getByRole('button'),
+    Object.assign(
```
illright, line 28: praise: this is some intense logic there, and I appreciate the research

issue: However, from what I discovered about JSDOM and the Testing Library is that testing drag is not a good idea in JSDOM, since it doesn't have layout at all. Maintainers of Testing Library suggest using Cypress for testing such stuff since it requires a real browser.

suggestion: perhaps we shouldn't cover this particular component with unit tests from Jest?

References:
- https://github.com/testing-library/user-event/issues/440#issuecomment-685010755 (This Kent C. Dodds guy is probably the most respectable guy in terms of testing from my current perspective)
- https://github.com/testing-library/user-event/issues/871 (An issue I created for the Testing Library when I was testing the scrubbing bar)

illright (changes requested):

src/lib/widgets/now-playing/specs/swipe-down-bar.spec.ts
VanishMax: Yeah, I've read these too, but couldn't understand what "no layout" actually mean. I'd appreciate you explaining the concept, and will easily remove the test

illright: Well, in simple terms, there's no "coordinate system", elements don't actually have a width and height. Everything is just collapsed into a single dot. That's how I understand it, anyway

illright: Which is why writing code that involves distances is risky and flaky in JSDOM

VanishMax: Removed the test

illright (changes requested):
Sorry for not mentioning it earlier, but this is actually a little more involved of a task. According to the acceptance criteria, the panel should follow your finger as you drag it around, not just minimize on a swipe-down gesture.

illright (approved):

PR #152, opened 2022-03-25
Cosmetic fixes

illright:
This PR will turn our app from a hot mess into an actually usable and pleasant-looking app. Compare with the current deployment on `main` and see for yourself

### DIFF
--- a/src/lib/widgets/now-playing/lib/lock-scroll-while-mounted.ts
+++ b/src/lib/widgets/now-playing/lib/lock-scroll-while-mounted.ts
@@ -1,0 +1,7 @@
+export function lockScrollWhileMounted() {
+  document.body.style.overflow = 'hidden';
+
+  return () => {
+    document.body.style.overflow = 'visible';
+  };
+}
--- a/src/lib/widgets/now-playing/ui/now-playing.svelte
+++ b/src/lib/widgets/now-playing/ui/now-playing.svelte
@@ -1,10 +1,9 @@
 <script lang="ts">
-  import { fly } from 'svelte/transition';
+  import { onMount } from 'svelte';
+  import { fly, fade } from 'svelte/transition';
   import IconArrowDown from '~icons/ic/round-keyboard-arrow-down';
-  import { IconButton } from '$lib/shared/ui';
-  import EpisodeDisplay from './episode-display.svelte';
   import { ToggleTranscript, Transcript } from '$lib/features/read-transcript';
-  import type { Episode, Podcast } from '$lib/shared/api';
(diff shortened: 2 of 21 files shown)


src/lib/widgets/now-playing/lib/lock-scroll-while-mounted.ts
```diff
@@ -0,0 +1,7 @@
+export function lockScrollWhileMounted() {
+  document.body.style.overflow = 'hidden';
+
+  return () => {
+    document.body.style.overflow = 'visible';
```
aabounegm, line 5: Why is it not `auto`?

khaledismaeel (commented):
Good job!

src/lib/widgets/now-playing/ui/now-playing.svelte
```diff
@@ -30,51 +27,73 @@
...
-  <div class="bg-slate-800 flex flex-col rounded-t-2xl py-10 px-8 items-center">
-    <div class="flex flex-col items-center relative">
-      {#if transcriptShown}
-        <Transcript {episode} class="absolute top-0 left-0 w-full h-full bg-slate-800 px-0.5" />
-      {/if}
-      <div class="flex flex-col items-center" class:invisible={transcriptShown}>
-        <EpisodeDisplay {episode} {podcast} />
+{#if $currentlyPlayingEpisode !== null}
+  <div
+    class="bg-black bg-opacity-20 flex flex-col justify-end fixed bottom-0 left-0 h-screen"
```
aabounegm, line 36: Is there a way to center it in the screen instead of aligning it to `left-0`?

aabounegm (commented):
Thanks for the amazing effort, it looks a lot more pleasant now.
Is it possible to also remove the horizontal scroll-bar in the episode transcript?

src/lib/widgets/now-playing/lib/lock-scroll-while-mounted.ts
illright: That's the default, I believe

src/lib/widgets/now-playing/ui/now-playing.svelte
illright: What for?

illright:
@aabounegm this is by no means the last PR for cosmetics, so we'll come back to the changes you wanted at a later stage

src/lib/widgets/now-playing/ui/now-playing.svelte
aabounegm: Because this looks ugly:
image

aabounegm (approved):

PR #162, opened 2022-03-28
Test organization

illright:


illright:
Ironic how a PR that adds tests actually decreases the coverage. It's because Codecov doesn't care which files we include in the coverage and uses its own algorithm to compute coverage. According to its algorithm, new files are tracked now

src/lib/entities/episode/model/specs/currently-playing-store.spec.ts
```diff
@@ -0,0 +1,6 @@
+import { currentlyPlayingEpisode } from '../..';
+
+it('conforms to the Svelte writable store contract', () => {
+  expect(currentlyPlayingEpisode.subscribe).toBeDefined();
+  expect(currentlyPlayingEpisode.set).toBeDefined();
+});
```
VanishMax, line 6: Question out of curiosity: what value does this test bring?

VanishMax (approved):
Very good! I checked the project organization as a whole and saw how consistent it is. Amazing work of your and a team in conforming to the FSD!

src/lib/entities/episode/model/specs/currently-playing-store.spec.ts
illright: Well, if we ever want to write our custom store or do any other changes, we'd wanna make sure we still follow the store contract.

I agree that, as it stands, there's little sense in testing a single line that you can just read. But it's a test nonetheless, and, being also a spec, it imposes restrictions on the contents of this file. In other words, it would make more sense if this test was written before the file itself, but we're not doing TDD

aabounegm: But wouldn't this be enforced by TypeScript?

illright: Yeah, it probably would. I just think it doesn't hurt to have these tests around

aabounegm: It also doesn't hurt to test that `1 + 1 = 2`, but that doesn't mean we should do it)

illright: @aabounegm would you like me to pull back these tests?

aabounegm: I prefer so, but I'll leave that decision to you. I just feel they are redundant as they do the work already done by the type checker

PR #166, opened 2022-03-29
Performance testing

### LINKED ISSUE #159: Set up performance testing with Lighthouse CI
You can use [this pre-built GitHub Action](https://www.foo.software/docs/lighthouse-check-github-action/intro)

**Acceptance criteria**:
- [Lighthouse CI](https://github.com/GoogleChrome/lighthouse-ci) is set up to run on each PR and send the results as a comment
- Lighthouse reports are available as artifacts to the workflow run
- If our scores are good enough, configure reasonable thresholds to fail the build if it's not met

VanishMax:
This PR closes #159.

Here's what happens here: on every PR and push to `main`, performance testing with Lighthouse will happen right after the E2E testing.  It will start the server and go through performance, accessibility, SEO and PWA metrics. 

This works pretty well, but after this merge the workflow will surely fail, because we don't know the results of Cast yet. I've added the `present: 'lighthouse:recommended'`, which requires 90%+ of performance. At the Lighthouse CI website, they recommend adding this preset and ticking off what our website fail. We don't know what it fails yet, so the second PR will be needed, I think.

You can check what it shows in the GitHub action run here: https://github.com/VanishMax/some-lighthouse-test/runs/5737478411


.github/actions/unit-tests-action/action.yml
```diff
@@ -59,3 +62,9 @@ runs:
         # we have already installed all dependencies above
         install: false
         cache-key: ${{ runner.os }}-node-${{ hashFiles('**/pnpm-lock.yaml') }}
+
+    - name: Performance test with Lighthouse 🧪
+      shell: bash
```
illright, line 67: question (non-blocking): why is this necessary?

lighthouserc.json
```diff
@@ -0,0 +1,18 @@
+{
+  "ci": {
+    "collect": {
+      "startServerCommand": "pnpm preview",
+      "startServerReadyPattern": "SvelteKit",
```
illright, line 5: question: why this pattern? it confused me a little bit because I didn't understand what this option should do until I looked it up.

suggestion: maybe we could use the `http://localhost:3000` as the pattern to ensure that this is the address that the server is on?

lighthouserc.json
```diff
@@ -0,0 +1,18 @@
+{
+  "ci": {
+    "collect": {
+      "startServerCommand": "pnpm preview",
```
illright, line 4: question: doesn't that only work after you built the server? or are you relying on Cypress to build the application?

suggestion: how about we move it into a separate workflow and ensure the application is built? The dependency on Cypress is kind of implicit and flaky

package.json
```diff
@@ -10,6 +10,7 @@
     "test:e2e": "cypress run",
     "test:e2e-open": "cypress open",
     "test:unit": "jest src",
+    "test:performance": "lhci autorun",
```
illright, line 13: praise: appreciate the clean integration into our patterns)

illright (changes requested):
I think the build is now failing due to conflicts of the Cypress runner and the Lighthouse runner. +1 reason to separate those into two different workflows

.github/actions/unit-tests-action/action.yml
VanishMax: IDE linters require it. I know that this could be removed, but my code will all become highlighted red 🤷‍♂️

VanishMax:
Yay, @illright, you genious comment in the issue worked!!! Check the tests failing: https://github.com/aabounegm/cast/runs/5749861233?check_suite_focus=true

What should we do with those tests now?

illright:
I guess we should make an issue to fix the performance problems, and merge this with a failing pipeline. However, I really don't like the 60 second sleep. We should do something more reliable in that area. Could you look into that, and create that issue, please?

VanishMax:
As I mentioned in the issue, there's a pretty cool GitHub Action that connects to the CloudFlare account API and wait for the preview app to be deployed. If the [issue that I made there](https://github.com/zentered/cloudflare-preview-url/issues/40) gets resolved, we could switch to using it real fast.

VanishMax:
Added a #167 issue for fixing performance testing

illright (approved):

VanishMax:
@illright How do we merge it without passing checks?

illright:
I believe it's not required to merge. You just need to merge `main` into this branch and you should be good to go

PR #170, opened 2022-04-01
Set up security headers to be read by Cloudflare

illright:
[Here's the documentation of Cloudflare Pages on headers](https://developers.cloudflare.com/pages/platform/headers/#:~:text=To%20attach%20headers%20to%20Cloudflare,root%20directory%20of%20the%20repository.)

[Here's why our current headers are unsatisfactory](https://securityheaders.com/?q=https%3A%2F%2Fcast-iu.pages.dev&followRedirects=on)

Related to #161

Issue #161, opened 2022-03-28
Run the app through the OWASP security checklist


illright:
It's a special challenge to find it. There are no defined acceptance criteria for this one, apart from just — _write up in the README whatever you want to say about this experience_. Give a link to the checklist, maybe write some code, idk.

illright:
The closest I could find to the mysterious _OWASP Security Checklist_ is the [OWASP Top Ten Project](https://owasp.org/www-project-top-ten/).

- [ ] [A01:2021-Broken Access Control](https://owasp.org/Top10/A01_2021-Broken_Access_Control/) — **blocked by not knowing how to authenticate in Supabase through Cypress**  
  (https://github.com/supabase/supabase/discussions/6177)
- [x] [A02:2021-Cryptographic Failures](https://owasp.org/Top10/A02_2021-Cryptographic_Failures/)
  - We don't store or manage data ourselves, relying on Supabase and HTTPS to communicate with it
  - We don't host our application ourselves, relying on Cloudflare and HTTPS to server the website
  - We don't generate our own keys for Supabase, relying on their own algorithms to generate strong keys
  - We use GitHub OAuth and store user metadata in an unexposed table in Supabase
- [ ] [A03:2021-Injection](https://owasp.org/Top10/A03_2021-Injection/)
- [x] [A04:2021-Insecure Design](https://owasp.org/Top10/A04_2021-Insecure_Design/)
  - We rely on third-party infrastructure and authentication protocols
- [ ] [A05:2021-Security Misconfiguration](https://owasp.org/Top10/A05_2021-Security_Misconfiguration/)
- [ ] [A06:2021-Vulnerable and Outdated Components](https://owasp.org/Top10/A06_2021-Vulnerable_and_Outdated_Components/)
- [x] [A07:2021-Identification and Authentication Failures](https://owasp.org/Top10/A07_2021-Identification_and_Authentication_Failures/)
  - GitHub manages authentication for us
- [ ] [A08:2021-Software and Data Integrity Failures](https://owasp.org/Top10/A08_2021-Software_and_Data_Integrity_Failures/)
- [ ] [A09:2021-Security Logging and Monitoring Failures](https://owasp.org/Top10/A09_2021-Security_Logging_and_Monitoring_Failures/)
- [ ] [A10:2021-Server-Side Request Forgery](https://owasp.org/Top10/A10_2021-Server-Side_Request_Forgery_%28SSRF%29/)

PR #171, opened 2022-04-01
Hide stack traces

illright:
Also this will finish the adoption of the `app` layer in FSD and fix the CSP that I forgot to fix in #170. Also related to #161


src/lib/app/lib/security-headers.ts
```diff
@@ -6,7 +6,6 @@ const securityHeaders = new Map([
   ['X-Frame-Options', 'SAMEORIGIN'],
   ['X-Content-Type-Options', 'nosniff'],
   ['Referrer-Policy', 'no-referrer'],
-  ['Content-Security-Policy', "script-src 'self'; frame-ancestors 'none';"],
```
VanishMax, line 9: Question: why do you use `new Map()` instead of simple object creation?

src/lib/pages/podcast-gallery/lib/load-podcasts.ts
```diff
@@ -0,0 +1,15 @@
+import type { Load } from '@sveltejs/kit';
+import { podcastList, podcasts } from '$lib/entities/podcast';
+
+export const loadPodcasts: Load = async () => {
```
VanishMax, line 4: Question: shouldn't this file be in the `api` slice instead of `lib`?

src/routes/podcasts/[id].svelte
```diff
@@ -1,37 +1,15 @@
...
 
-  if (podcast) {
-    podcastStore.update((podcasts) => {
-      if (!podcasts.some((p) => p.id === podcast!.id)) {
-        podcasts.push(podcast!);
-      }
-      return podcasts;
-    });
-  }
+  addPodcastToStore(podcast);
```
Comment (VanishMax) [inline src/routes/podcasts/[id].svelte:12]: Question: what does adding a podcast to the list of podcasts do?

VanishMax (approved):
Praise: you've done a great job with ensuring security and adding errors cathing in the app! I am really curious how it all works, so I left a couple of question comments to the PR and asking, if it is possible, for a video or a call with more detailed explanation on how the security is checked.

src/lib/app/lib/security-headers.ts
illright: I just don't really trust objects at all. You know how iterating over an object might give you some unexpected properties? I don't know if that really happens or not, I just dislike objects for that reason. Maps are a natural way of storing key-value pairs, and they are rock solid, so I just like them much more

src/lib/pages/podcast-gallery/lib/load-podcasts.ts
illright: Good call, I think it should

src/routes/podcasts/[id].svelte
Comment (illright) [reply src/routes/podcasts/[id].svelte:12]: It caches that podcast's metadata to avoid downloading it again later

illright:
Responding to your call — yeah, sure, I can record a demo about the whole security stuff that I've done

PR #172, opened 2022-04-02
Fuzz testing for IDs in the URL

### LINKED ISSUE #160: Build fuzz tests
I imagine we should do them with Cypress, going over the [Big List of Naughty Strings](https://www.npmjs.com/package/blns) and using them in place of various fields like usernames, podcast and episode names, authors, transcripts (!!!).

**Acceptance criteria**:
- a fuzz testing spec is added to run with Cypress
- this spec is tagged to avoid running it every time
- the CI command is updated not to run this spec when running Cypress tests

illright:
Closes #160, related to #161

I decided not to fuzz-test anything else because that's really the only place where untrusted input may come from


src/lib/pages/podcast-episodes/api/load-podcast-from-url.ts
```diff
@@ -3,7 +3,12 @@ import { podcastGet } from '$lib/entities/podcast';
 
 export const loadPodcastFromURL: Load = async ({ params }) => {
   try {
-    const podcast = await podcastGet(parseInt(params.id, 10));
+    const integerID = parseInt(params.id, 10);
+    if (isNaN(integerID)) {
+      throw integerID;
```
aabounegm, line 8: `throw`ing `NaN` feels a bit lousy, tbh)
Wouldn't a 404 error be more appropiate?

illright: Idk, I just figured that if we were to return 404s, we'd better design a decent 404 experience, but in this way we just redirect to Podcast Gallery

alkaitagi (approved):

PR #174, opened 2022-04-04
Connect history & likes to backend

### LINKED ISSUE #110: Implement the synchronization of favorites and history with the backend
Blocked by #109 and #100.

**Acceptance criteria**:
- upon loading the app, if a valid token is present, synchronization should happen
- upon marking/unmarking an episode as favorite, the change should be communicated to the backend

alkaitagi:
Closes #110.


src/lib/features/listening-history/model/store.ts
```diff
@@ -12,16 +26,36 @@ export const listeningHistory = persistentWritable<number[]>([], {
...
   listeningHistory.update((history) => {
     return [podcastId, ...history.filter((id) => id !== podcastId)].slice(0, 6);
   });
 };
 
-currentlyPlayingEpisode.subscribe(($episode) => {
+/**
+ * Function that add a row into history SB table.
+ *
+ * @param podcastId {string} – an id to add to history
```
illright, line 40: nitpick (non-blocking): this type in the JSDoc is incorrect

suggestion: let's just remove this entire line, it doesn't add clarity to the comment

src/lib/features/listening-history/index.ts
```diff
@@ -1 +1,4 @@
-export { addToListeningHistory, listeningHistory } from './model/store';
+export {
+  addToLocalListeningHistory as addToListeningHistory,
```
illright, line 2: question: why do we only expose the "local" version of this function? won't this lead to unsynchronization?

src/lib/features/listening-history/model/store.ts
```diff
@@ -12,16 +26,36 @@ export const listeningHistory = persistentWritable<number[]>([], {
...
+ */
+export const addToCloudListeningHistory = async (podcastId: number) => {
+  const u = get(user);
+  if (!u) return;
+  await supabaseClient.from('history').insert([
+    {
+      // eslint-disable-next-line camelcase
+      podcast_id: podcastId,
+      // eslint-disable-next-line camelcase
+      user_id: u.id,
```
illright, line 50: suggestion (non-blocking): looks like there's a way to let Supabase figure out the user ID itself: https://github.com/supabase/supabase/discussions/4368#discussioncomment-1769320

illright:
Looks good so far, but since this PR will close the issue, I'll hold on to approving until you implement the like sync as well

alkaitagi:
@illright I understand my current solution is not very performant & scalable. But at least it works, and I might continue optimizations in this PR.

alkaitagi:
Oh well, the tests are broken too.

src/lib/pages/library/ui/library-page.svelte
```diff
@@ -5,8 +5,10 @@
   import { UserContainer } from '$lib/entities/user';
   import { SignInButton, SignOutButton } from '$lib/features/authenticate';
   import type { Episode } from '$lib/shared/api';
+  import { getEpisodes } from '../lib/load-episodes';
 
-  $: episodes = Array.from<Episode>($likesStore.values());
+  let episodes: Episode[] = [];
+  $: getEpisodes([...$likesStore]).then((r) => (episodes = r));
```
illright, line 11: thought: I see you using this pattern quite frequently, I'd like to suggest a more conventional Svelte alternative — the [`{#await}` block](https://svelte.dev/tutorial/await-blocks), just in case you didn't know about it

src/lib/pages/library/lib/get-episode.ts
```diff
@@ -0,0 +1,12 @@
+import { supabaseClient } from '$lib/shared/api';
```
illright, line 1: suggestion: I think this file belongs on the `entities` layer

src/lib/pages/library/ui/library-page.svelte
```diff
@@ -4,9 +4,9 @@
   import { EpisodeCard } from '$lib/widgets/episode';
   import { UserContainer } from '$lib/entities/user';
   import { SignInButton, SignOutButton } from '$lib/features/authenticate';
-  import type { Episode } from '$lib/shared/api';
+  import { getEpisode } from '../lib/get-episode';
 
-  $: episodes = Array.from<Episode>($likesStore.values());
+  const liked = [...$likesStore];
```
illright, line 9: issue: this will not be updated as the `likesStore` updates

suggestion: use a reactive declaration instead:
```suggestion
  $: liked = [...$likesStore];
```

src/lib/pages/library/ui/library-page.svelte
```diff
@@ -22,8 +22,17 @@
 <!-- Favorites -->
 <section class="flex flex-col p-4 gap-4 items-stretch" aria-labelledby="favorites-header">
   <H1 id="favorites-header">Favorites</H1>
-  {#each episodes as episode (episode.id)}
-    <EpisodeCard {episode} />
+
+  {#each liked as id}
```
illright, line 26: suggestion (non-blocking): use a keyed each block, that way the updates to the likes will be more efficient
```suggestion
  {#each liked as id (id)}
```

src/lib/features/listening-history/model/store.ts
```diff
@@ -12,14 +26,31 @@ export const listeningHistory = persistentWritable<number[]>([], {
...
     return [podcastId, ...history.filter((id) => id !== podcastId)].slice(0, 6);
   });
 };
 
+/**
+ * Function that adds a podcast id into `history` SB table.
+ */
+const addToCloudListeningHistory = async (podcastId: number) => {
+  if (!get(user)) return;
+  await supabaseClient.from('history').insert([
```
illright, line 40: issue: this is an unhandled promise, it might throw, for example, when we are offline

suggestion: remove the check for the user above and just wrap the `await` in a `try-catch` to prevent both unauthenticated requests and any network errors

src/lib/features/listening-history/model/store.ts
```diff
@@ -1,8 +1,22 @@
 import { currentlyPlayingEpisode } from '$lib/entities/episode';
+import { supabaseClient, type Episode } from '$lib/shared/api';
+import { user } from '$lib/entities/user';
 import { localStorageAdapter, persistentWritable } from 'svelte-persistent-writable';
+import { get } from 'svelte/store';
+
+user.subscribe(async ($user) => {
+  if (!$user) return;
+  const hist = await supabaseClient
```
illright, line 9: suggestion (non-blocking): I think this code should go in the `api` segment. Also the code below that adds to the `history` table

src/lib/features/listening-history/model/specs/store.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+  episodeNumber: 1,
+  podcastID: 1,
+  title: 'Episode 1',
+  audioUrl: 'http://example.com',
+  duration: 5,
+};
+
+const expectedHistory: number[] = [exampleEpisode.id];
+
+it('listen to an episode', () => {
```
illright, line 16: suggestion (non-blocking): the code reads better when using -s forms:
```suggestion
it('listens to an episode', () => {
```

but then
question: how does this test that the code "listens to an episode"?

src/lib/features/listening-history/model/specs/store.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+  addToListeningHistory(exampleEpisode);
+  expect(get(listeningHistory)).toEqual(expectedHistory);
```
illright, line 18: issue: this test doesn't test the network interactions, which I think are an important aspect

src/lib/features/listening-history/model/store.ts
illright: this also applies to several other places in the code

src/lib/features/like-episode/model/like.ts
```diff
@@ -1,39 +1,48 @@
-import type { Episode } from '$lib/shared/api';
+import { supabaseClient, type Episode } from '$lib/shared/api';
 import { persistentWritable, localStorageAdapter } from 'svelte-persistent-writable';
+import { user } from '$lib/entities/user';
+import { get } from 'svelte/store';
+
+user.subscribe(async ($user) => {
+  if (!$user) return;
+  const hist = await supabaseClient
```
illright, line 8: issue (non-blocking): copy-pasted variable names are a bit confusing :)

illright (changes requested):

src/lib/features/like-episode/model/like.ts
```diff
@@ -30,19 +30,28 @@ export const likesStore = persistentWritable(new Set<number>(), {
...
 
 async function deleteCloudLike(episodeId: number) {
-  const u = get(user);
-  if (!u) return;
-  await supabaseClient.from('favourites').delete().eq('user_id', u.id).eq('episode_id', episodeId);
+  try {
+    await supabaseClient
+      .from('favourites')
+      .delete()
+      .eq('user_id', get(user)?.id)
```
illright, line 50: question: is this actually necessary? I think that RLS in Supabase, when configured properly for reading, will only give you the records for the current user anyway, regardless of whether you filter it or not

src/lib/features/like-episode/model/like.ts
```diff
@@ -30,19 +30,28 @@ export const likesStore = persistentWritable(new Set<number>(), {
...
-  ]);
+  try {
+    await supabaseClient.from('favourites').insert([
+      {
+        // eslint-disable-next-line camelcase
+        episode_id: episodeId,
+      },
+    ]);
+  } catch (e) {
+    console.error(e);
```
illright, line 41: issue: I would appreciate if we kept on hiding stack traces, continuing on the work I've done in https://github.com/aabounegm/cast/pull/171.

suggestion: here I would suggest propagating the error to the caller to let them handle it however they see fit

src/lib/features/listening-history/model/specs/store.spec.ts
```diff
@@ -13,7 +13,7 @@ const exampleEpisode: Episode = {
 
 const expectedHistory: number[] = [exampleEpisode.id];
 
-it('listen to an episode', () => {
+it('add episode to the history', () => {
```
illright, line 16: nitpick (non-blocking): I like to think of this function name `it()` as a part of the sentence, so to be grammatically correct, we would say "it adds an episode..."
```suggestion
it('adds an episode to the history', () => {
```

src/lib/pages/library/ui/library-page.svelte
```diff
@@ -4,9 +4,9 @@
   import { EpisodeCard } from '$lib/widgets/episode';
   import { UserContainer } from '$lib/entities/user';
   import { SignInButton, SignOutButton } from '$lib/features/authenticate';
-  import { getEpisode } from '../lib/get-episode';
+  import { fetchEpisode } from '$lib/entities/episode/api/fetch-episode';
```
illright, line 7: issue: according to FSD limitations, we're not allowed to dig into the structure of another slice

illright (changes requested):

src/lib/pages/library/ui/library-page.svelte
alkaitagi: Actually this was intented. Like I said it during my demo, I want te episodes to still be on UI until reload, in case user changes their mind.

illright: oh, alright then, I take it back

src/lib/features/listening-history/model/specs/store.spec.ts
alkaitagi: Added podcast fetching tests in `src\lib\entities\episode\api\specs\get-episode.spec.ts`. Let me know if that's not sufficient.

src/lib/entities/episode/api/get-episode.ts
```diff
@@ -0,0 +1,12 @@
+import { supabaseClient } from '$lib/shared/api';
+import { transformEpisodeRequest, type SBEpisode } from '$lib/shared/api/adapters';
```
illright, line 2: issue: again, a little violation of FSD, but I understand why. This stuff is not exported from that module

suggestion: let's just export it and import it in a valid way

src/lib/features/like-episode/model/like.ts
```diff
@@ -1,39 +1,57 @@
...
-    likes.delete(episode.id);
-    return likes;
-  });
+async function deleteCloudLike(episodeId: number) {
+  try {
+    await supabaseClient.from('favourites').delete().eq('episode_id', episodeId);
+  } catch {
+    return {
+      status: 500,
+      error: 'Server error. Check internet connection and ensure you are signed in.',
```
illright, line 52: question: where is this handled/displayed?

src/lib/features/listening-history/model/store.ts
illright: You resolved this conversation without a response, what does that mean?

src/lib/features/listening-history/model/specs/store.spec.ts
illright: It's sufficient for testing `getEpisode`, but in this thread I was talking about the `addToListeningHistory` function, namely, its logic of communicating with the backend.

src/lib/features/listening-history/model/specs/store.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+  episodeNumber: 1,
+  podcastID: 1,
+  title: 'Episode 1',
+  audioUrl: 'http://example.com',
+  duration: 5,
+};
+
+const expectedHistory: number[] = [exampleEpisode.id];
+
+it('adds an episode to the history', () => {
```
illright, line 16: suggestion (non-blocking): I think this test suite should also test the subscriptions — that the history is fetched on login and that playing an episode adds it to history. However, this expands from unit tests into integration tests slightly, so do it if you see fit, we might cover it in large-scale E2E tests later anyway

illright (changes requested):

src/lib/features/like-episode/model/like.ts
alkaitagi: Added `alert(e.error)` calls.

src/lib/features/listening-history/model/store.ts
alkaitagi: Probably confused it with another comment. Thank for pointing out.

alkaitagi: Done. Please, re-review this point. I hope to add corresponding tests soon.

src/lib/features/listening-history/model/specs/store.spec.ts
alkaitagi: Added more tests at `src\lib\features\listening-history\api` and `src\lib\features\like-episode\api`.

illright (approved):
praise: thanks for withstanding this excruciating review process :)

issue (non-blocking): I'm not a fan of these `alert()` calls, but let's let Khaled amend this in his PR about notificational snackbars

thought: I noticed you returned objects like `{ status: 500, error: '...' }`, I assume you mimicked the way we handled exceptions in [loader functions](https://kit.svelte.dev/docs/loading#output), but in your case the function is not a loader, so returning these codes serves no purpose. I took the liberty of removing them

PR #182, opened 2022-04-06
E2E testing auth user

### LINKED ISSUE #177: Create a testing user in Supabase and set it up for Cypress
I already messed around there with that stuff, trying to create a user without the password, but failed miserably. See https://github.com/supabase/supabase/discussions/6177 for details.

I think what we should do is enable email signups and create a testing user with a proper randomly generated password. We can then just change the field in the database that says the email is not confirmed to make it confirmed, and we should have a decent experience to test auth-protected paths of our app.

**Acceptance criteria**:
- A separate user for testing should be present in Supabase with a randomly generated 64+ character password
- a `cy.login()` command should be added that sets the testing environment up with an authenticated user using the provided credentials
- The generated password should be accepted as an env variable

aabounegm:
The PR is marked as draft since the second acceptance criterion isn't met.
Since we do not have access to the APIs (such as `supabaseClient`) in E2E tests, it is not possible to login in a way the user cannot. In other words, we need a login form with email/password fields for the users.
Suggestions are appreciated.

Closes #177


illright:
I'm not sure why you're saying that we don't have access to the Supabase client. The library should be available to Cypress in the same way that is is available to SvelteKit, I imagine it's a matter of just creating a client object again

cypress/support/login.ts
```diff
@@ -0,0 +1,13 @@
+import { createClient } from '@supabase/supabase-js';
+
+export default function login() {
+  const appUrl: string = Cypress.env('supabaseAppUrl');
+  const publicAnonKey: string = Cypress.env('supabasePublicAnonKey');
+  const supabase = createClient(appUrl, publicAnonKey);
+
+  const email = 'testing@cast-iu.pages.dev';
```
illright, line 8: issue (non-blocking): might have been better not to hard-code this e-mail, but it's not a big deal

illright (approved):

cypress/support/login.ts
aabounegm: I thought the same, but couldn't figure out where to store it instead. Having it in `.env` would be inconvenient since it will have to be added to all of our environments (including CI) when it's not really a secret. I guess I can just add it to `cypress.json`, since it's not even worth creating a separate `cypress.env.json` just for this one variable

Issue #167, opened 2022-03-30
Fix performance testing


VanishMax:
For now, performance testing uses `present: 'lighthouse:recommended'`, which sets the recommended configurations to the `lighthouserc.json`. Our application cannot handle them for now, thus [the CI fails](https://github.com/aabounegm/cast/runs/5750907336?check_suite_focus=true).

The task is to fix these issues or configure the `assess` field, so the CI will pass successfully.

Acceptance criteria:
* CI passes
* Problematic points are fixed or the threshold for them is changed in the `assess` field of the Lighthouse config. Also, it would be nice to write a description in this PR of why the threshold has been changed.

VanishMax:
One more thing to do here is to check the checkbox in our Readme about Lighthouse tests

illright:
That's done already

VanishMax:
Wait, is it different for us?

<img width="688" alt="image" src="https://user-images.githubusercontent.com/29180358/161680244-b8ca48ac-6274-4d8b-b622-169afec58b0a.png">

illright:
Oh, sorry, forgot that it is still not merged

PR #184, opened 2022-04-07
Configure auto-generation for meta images

### LINKED ISSUE #183: Fix the icon assets for PWA
On Android, the icon looks pixelated. Also, on the splash screen, it's very tiny. Maybe we could set up some sort of automatic build process to create images of required resolution (but I don't think that's highly probable)

illright:
Closes #183, related to #167

I had **such a blast** configuring this, wonderful DX


illright:
@aabounegm there was actually an issue in the installed PWA, I patched it up for now, please, test it on your phone. @VanishMax, I'd also like you to test it on the iPhone

src/lib/app/meta/pwa-manifest.ts
```diff
@@ -0,0 +1,52 @@
...
+      type: 'image/png',
+      purpose: 'any',
+      sizes: '192x192',
+    },
+    {
+      src: appIconPng512,
+      type: 'image/png',
+      purpose: 'maskable',
+      sizes: '512x512',
+    },
```
VanishMax, line 40: Question: why do these 2 blocks have different purpose?

VanishMax (approved):
Praise: I'm in love with this piece of code.

aabounegm:
It looks much better now. The splash screen has the correct colors and image size, and the app icon is no longer pixelated.

https://user-images.githubusercontent.com/11016151/162384753-378212be-7cbb-4ead-a0e2-36c7b8ebc4a5.mp4

src/lib/app/meta/pwa-manifest.ts
illright: I just figured it's best to have a maskable image as well as a non-maskable one so that the browsers that prefer not to mask would have a choice. Just a random guess, honestly

PR #185, opened 2022-04-07
Harden the CSP

illright:
Related to #167. For information, see https://web.dev/csp-xss/


VanishMax (approved):
I bet you are a pro in web security now 😎

PR #199, opened 2022-04-12
Security tests

### LINKED ISSUE #161: Run the app through the OWASP security checklist

illright:
Closes #161 

The final piece — E2E tests for API security

### DIFF
--- a/cypress/integration/access-control/history.spec.ts
+++ b/cypress/integration/access-control/history.spec.ts
@@ -1,0 +1,126 @@
+/* eslint-disable camelcase */
+import type { SBHistoryEntry, SBPodcast } from '$lib/shared/api/adapters';
+
+describe('For anonymous users', () => {
+  it('forbids reading', () => {
+    cy.createSupabaseClient().then(async (supabase) => {
+      const response = await supabase.from<SBHistoryEntry>('history').select('*').limit(1).single();
+      expect(response.status).to.be.gte(400).and.be.lessThan(500);
+    });
+  });
+
+  it('forbids creating', () => {
+    cy.createSupabaseClient().then(async (supabase) => {
+      const response = await supabase.from<SBHistoryEntry>('history').insert({ podcast_id: 1 });
+      expect(response.status).to.be.gte(400).and.be.lessThan(500);
+    });
+  });
+});
+
+describe('For authorized users', () => {
+  it('permits reading own rows', () => {
+    cy.login().then(async ([auth, supabase]) => {
+      const userID = (await auth).user?.id;
+      if (userID !== undefined) {
+        const response = await supabase.from<SBHistoryEntry>('history').select('user_id');
+        expect(response.status).to.be.gte(200).and.be.lessThan(300);
+        if (response.data !== null) {
+          for (const row of response.data) {
+            expect(row.user_id).to.be.equal(userID);
+          }
+        } else {
+          expect(response.data).not.to.be.null;
+        }
+      }
+    });
+  });
+
+  it('forbids reading rows of other users', () => {
+    cy.login().then(async ([auth, supabase]) => {
+      const userID = (await auth).user?.id;
+      if (userID !== undefined) {
+        const response = await supabase
+          .from<SBHistoryEntry>('history')
+          .select('*')
+          .filter('user_id', 'neq', userID)
+          .single();
+        expect(response.status).to.be.gte(400).and.be.lessThan(500);
+      }
+    });
+  });
+
+  it('permits creating rows with your own user ID', () => {
+    cy.login().then(async ([auth, supabase]) => {
+      const userID = (await auth).user?.id;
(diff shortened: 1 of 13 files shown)


cypress/support/login.ts
```diff
@@ -1,13 +1,11 @@
-import { createClient } from '@supabase/supabase-js';
+import createSupabaseClient from './create-supabase-client';
```
aabounegm, line 1: Why is it imported explicitly here instead of calling `cy.createSupabaseClient`?

src/lib/shared/api/adapters.ts
```diff
@@ -24,10 +24,23 @@ export interface SBEpisode {
 }
 
 export interface SBTranscript {
+  id: number;
   episode_id: SBEpisode['id'];
   content: string;
 }
 
+export interface SBFavourite {
```
aabounegm, line 32: I, for one, would almost always prefer the British spelling, except that we already used `favorite` everywhere else in the project, and I value consistency more)

src/lib/shared/api/adapters.ts
```diff
@@ -69,3 +82,7 @@ export const transformPodcastRequest = (podcast: SBPodcast): Podcast | null => {
       .sort((e1, e2) => e1.episodeNumber - e2.episodeNumber),
   };
 };
+
+export const transformFavourite = (favourite: SBFavourite): Episode['id'] => {
```
aabounegm, line 86: I don't see this function being used anywhere. I also don't feel `(f) => f.episode_id` is worth defining a dedicated function unless it's used more than once. Plus, the name doesn't really convey what it actually does

cypress/integration/access-control/history.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+      const userID = (await auth).user?.id;
+      if (userID !== undefined) {
+        const response = await supabase.from<SBHistoryEntry>('history').select('user_id');
+        expect(response.status).to.be.gte(200).and.be.lessThan(300);
+        if (response.data !== null) {
+          for (const row of response.data) {
+            expect(row.user_id).to.be.equal(userID);
+          }
+        } else {
+          expect(response.data).not.to.be.null;
```
aabounegm, line 32: Feels like it can be moved before the `if`, with a non-null assertion in the `for` loop, since it is obvious that this `expect` is never going to succeed

aabounegm (commented):
While these `access-control` tests feel more like "unit tests" for supabase rather than integration tests for our app, excellent job writing them in general

src/lib/shared/api/adapters.ts
illright: I have very flaky standards when it comes to British vs English, because I like to say _favorite_ yet _neighbour_, but in this case I think it was Magomed who named the table in Supabase, and I thought that we should keep the Supabase objects with a `u`, but the rest of the code may proceed without it. Because of that, I don't really want to change this to be without the `u`

cypress/integration/access-control/history.spec.ts
illright: Our linter is configured against non-null assertions

src/lib/shared/api/adapters.ts
illright: Ah, right, I meant to refactor some of Magomed's new code to use this function. It does convey what the function does — it transforms the Supabase representation of favourites to our representation of favourites (check me out, juggling `u`)

cypress/support/login.ts
illright: It won't work otherwise for some reason

aabounegm: Hmm, it worked for me. I removed the import and simply added `cy.` and all of the tests passed

cypress/integration/access-control/history.spec.ts
aabounegm: Well, you can add `if (response.data === null) return;` after the `expect` just for the flow analysis, but it would still make more sense than writing an `expect` that you know for sure will fail.

cypress/support/login.ts
illright: Let's keep it at that to avoid any possibilities of test flake

cypress/integration/access-control/history.spec.ts
illright: So what would you like me to do here?

cypress/support/login.ts
aabounegm: I don't see why it would be flaky, but fine

cypress/integration/access-control/history.spec.ts
aabounegm: Nothing. I did it myself. Hope you don't mind)

PR #200, opened 2022-04-14
Fix Lighthouse errors

### LINKED ISSUE #167: Fix performance testing

aabounegm:
I enabled the CSP again but disabled the `strict-dymanic` source since SvelteKit doesn't support it properly yet (https://github.com/sveltejs/kit/issues/3558).
As for the `unused-javascript` issue, I disabled the assertion since it is mostly not in our control (and SvelteKit basically closed the issue as `wontfix` - https://github.com/sveltejs/kit/issues/1371). It might be possible to work around it (according to [this Reddit comment](https://www.reddit.com/r/sveltejs/comments/tih8sx/pagespeed_lighthouse_reduce_unused_javascript/)), but it feels like a dirty hack, so I'd say it's good as it is now.

Resolves #167

### DIFF
--- a/src/app.html
+++ b/src/app.html
@@ -4,7 +4,6 @@
     <meta charset="utf-8" />
     <meta name="description" content="Listen to podcasts on the web, wherever!" />
     <link rel="manifest" href="%svelte.assets%/manifest.json" />
-    <link rel="preconnect" href="https://sklgqhbbbhticzdgegcj.supabase.co" />
     <meta name="theme-color" content="#334155" />
     <meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover" />
     %svelte.head%
--- a/src/app.html
+++ b/src/app.html
@@ -4,6 +4,7 @@
     <meta charset="utf-8" />
     <meta name="description" content="Listen to podcasts on the web, wherever!" />
     <link rel="manifest" href="%svelte.assets%/manifest.json" />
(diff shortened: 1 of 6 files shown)


src/app.html
```diff
@@ -4,6 +4,7 @@
     <meta charset="utf-8" />
     <meta name="description" content="Listen to podcasts on the web, wherever!" />
     <link rel="manifest" href="%svelte.assets%/manifest.json" />
+    <link rel="preconnect" href="https://sklgqhbbbhticzdgegcj.supabase.co" />
```
illright, line 7: issue: I don't think we should hard-code this

suggestion: in the `app` layer, there are a few components that manage meta tags, perhaps you could move it there and use the value from the environment variable?

illright (changes requested):
praise: thank you so much for taking this off of me and figuring out the CSP issue :)

illright (approved):

PR #201, opened 2022-04-15
Audio play fix

### LINKED ISSUE #155: Switching episode pauses audio
According to @aabounegm: "Playing another episode while one is already playing switches the current episode correctly, and leaves paused as false also correctly, but for some reason the audio element itself stops playing."

We need to fix this _and_ add unit tests to cover this behavior.

alkaitagi:
Closes #155.

### DIFF
--- a/src/lib/entities/audio/model/audio-instance.ts
+++ b/src/lib/entities/audio/model/audio-instance.ts
@@ -1,4 +1,4 @@
-import { get, writable } from 'svelte/store';
+import { get, writable, type Unsubscriber } from 'svelte/store';
 
 // Because apparently Svelte thinks it knows better how the Web APIs should be.
 export type SvelteTimeRanges = Array<{ start: number; end: number }>;
@@ -11,8 +11,18 @@ export const paused = writable(true);
 export const buffered = writable<SvelteTimeRanges>([]);
 
 export const play = (_src?: string) => {
-  if (_src) src.set(_src);
-  paused.set(false);
+  if (_src) {
+    pause();
+    duration.set(0);
+    let s: Unsubscriber | undefined = undefined;
+    s = duration.subscribe((v) => {
+      if (v) {
+        paused.set(false);
+        s?.();
+      }
+    });
+    src.set(_src);
+  } else paused.set(false);
 };
 
 export const pause = () => paused.set(true);
--- a/src/lib/entities/audio/model/audio-instance.ts
+++ b/src/lib/entities/audio/model/audio-instance.ts
@@ -1,4 +1,4 @@
-import { get, writable, type Unsubscriber } from 'svelte/store';
+import { get, writable } from 'svelte/store';
 
 // Because apparently Svelte thinks it knows better how the Web APIs should be.
 export type SvelteTimeRanges = Array<{ start: number; end: number }>;
@@ -14,15 +14,16 @@ export const play = (_src?: string) => {
   if (_src) {
     pause();
     duration.set(0);
-    let s: Unsubscriber | undefined = undefined;
-    s = duration.subscribe((v) => {
+    const unsub = duration.subscribe((v) => {
       if (v) {
         paused.set(false);
-        s?.();
+        unsub();
       }
     });
     src.set(_src);
-  } else paused.set(false);
+  } else {
+    paused.set(false);
+  }
 };
 
 export const pause = () => paused.set(true);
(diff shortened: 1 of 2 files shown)


src/lib/entities/audio/model/specs/audio-instance.spec.ts
```diff
@@ -41,4 +41,14 @@ describe('Audio playback API', () => {
     expect(get(currentTime)).toEqual(get(duration));
     pause();
   });
+
+  it('plays twice without pause', async () => {
+    playDefault();
+    await new Promise((r) => setTimeout(r, 1000));
```
aabounegm, line 47: Someone pointed in a previous PR that it's better to use `jest.advanceTimersByTime`, and that would require
```ts
jest.useFakeTimers();
jest.spyOn(global, 'setTimeout');
```
somewhere on the top. See [`$lib/shared/ui/specs/snackbar.spec.ts`](https://github.com/aabounegm/cast/blob/5cbc1fd9bf61dcdf5ca0b28d74875a6cf965ae80/src/lib/shared/ui/specs/snackbar.spec.ts) for example usage.

src/lib/entities/audio/model/audio-instance.ts
```diff
@@ -11,8 +11,18 @@ export const paused = writable(true);
 export const buffered = writable<SvelteTimeRanges>([]);
 
 export const play = (_src?: string) => {
-  if (_src) src.set(_src);
-  paused.set(false);
+  if (_src) {
+    pause();
+    duration.set(0);
+    let s: Unsubscriber | undefined = undefined;
+    s = duration.subscribe((v) => {
```
aabounegm, line 18: I don't think it needs to have `undefined` in its type since it never gets assigned to it. This can simply become
```suggestion
    const s = duration.subscribe((v) => {
```
and the optional chaining won't be needed

aabounegm: Also, a more meaningful name (such as `unsubscribe`) would be helpful

src/lib/entities/audio/model/audio-instance.ts
```diff
@@ -11,8 +11,18 @@ export const paused = writable(true);
...
+      if (v) {
+        paused.set(false);
+        s?.();
+      }
+    });
+    src.set(_src);
+  } else paused.set(false);
```
aabounegm, line 25: Can you use braces here as well just for consistency's sake?

aabounegm (changes requested):

src/lib/entities/audio/model/audio-instance.ts
```diff
@@ -11,8 +11,19 @@ export const paused = writable(true);
 export const buffered = writable<SvelteTimeRanges>([]);
 
 export const play = (_src?: string) => {
-  if (_src) src.set(_src);
-  paused.set(false);
+  if (_src) {
+    pause();
+    duration.set(0);
```
illright, line 16: issue: I don't quite understand this `if` block. Could you please write a comment here explaining why this stuff is needed?

illright (changes requested):

src/lib/entities/audio/model/specs/audio-instance.spec.ts
```diff
@@ -41,4 +44,14 @@ describe('Audio playback API', () => {
     expect(get(currentTime)).toEqual(get(duration));
     pause();
   });
+
+  it('plays twice without pause', async () => {
+    playDefault();
+    jest.advanceTimersByTime(1000);
```
illright, line 50: praise: this is great :)

src/lib/entities/audio/model/audio-instance.ts
alkaitagi: Setting `duration` to 0 and subscribing to it helps with catching the moment when the the URL takes effect thus "populatnig" the writables with data. And upon that moment, when the data is loaded and the player is ready, I call play.

alkaitagi: It is of course a hack. But the fact that the audio stops after URL change also seems to be a bug. So I hack the bug.

aabounegm (approved):

src/lib/entities/audio/model/audio-instance.ts
illright: I meant to add a comment to the source code for everyone to read in the future

aabounegm (approved):

illright (approved):

PR #203, opened 2022-04-16
Trending podcasts

### LINKED ISSUE #190: Implement the fetching of trending podcasts
**Acceptance criteria**:
- the backend should process the listening histories of all users securely
- a new view `trending_podcasts` should be created to provide IDs of the 6 most commonly listened podcasts
- the data from the backend should be wired up to the frontend
- unit tests should be added where necessary

aabounegm:
View created on Supabase using:
```sql
CREATE VIEW trending_podcasts AS
  SELECT podcast_id
  FROM history
  GROUP BY podcast_id
  ORDER BY COUNT(podcast_id) DESC
  LIMIT 6;
```

I couldn't really think of a unit test that can be written, due to how simple the code is (it is simply a wrapper around an API call).

Closes #190

### DIFF
--- a/src/lib/entities/podcast/ui/podcast-tile.svelte
+++ b/src/lib/entities/podcast/ui/podcast-tile.svelte
@@ -2,11 +2,27 @@
   import cx from 'clsx';
   import { Image } from '$lib/shared/ui';
   import type { Podcast } from '$lib/shared/api';
+  import { onMount } from 'svelte';
 
   let _class = '';
   export { _class as class };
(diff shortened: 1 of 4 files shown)


illright (approved):

aabounegm:
There are still some warnings in the performance testing that can be addressed, but we can tackle them later. I think the "Trending Podcasts" section in particular loads later than others because the "listening history" is stored locally, and "All podcasts" is fetched in the `load` function of the component context.

PR #204, opened 2022-04-16
Offline Listening E2E test

### LINKED ISSUE #195: Create E2E tests related to offline listening
See https://github.com/aabounegm/cast/issues/176#issuecomment-1088375282 for a list of requirements, section **Audio download and offline listening**. Figure out a way to simulate not being connected to the internet.

**Acceptance criteria**:

aabounegm:
Closes #195

### DIFF
--- a/src/lib/pages/podcast-gallery/api/load-podcasts.ts
+++ b/src/lib/pages/podcast-gallery/api/load-podcasts.ts
@@ -5,7 +5,8 @@ export const loadPodcasts: Load = async () => {
   try {
     podcasts.set(await podcastList());
     return {};
-  } catch {
+  } catch (e) {
+    console.error(e);
     return {
       status: 500,
       error:
--- a/cypress/fixtures/sample-podcasts.json
+++ b/cypress/fixtures/sample-podcasts.json
@@ -1,17 +1,24 @@
 {
   "id": 1,
-  "coverUrl": "https://influencermarketing.ai/wp-content/uploads/2020/05/Podcast-blog-6.jpg",
+  "coverArt": {
+    "name": "caucastalk.jpg"
+  },
   "title": "Bi Co Bo Co",
   "author": "Cciqhaqxa",
   "episodes": [
     {
       "id": 1,
-      "audioUrl": "https://www.soundhelix.com/examples/mp3/SoundHelix-Song-1.mp3",
+      "audio": {
+        "name": "CaucasTalk_Episode114.mp3"
+      },
       "title": "Ahulgox",
       "duration": 124
     },
     {
       "id": 2,
+      "audio": {
+        "name": "CaucasTalk_Episode113-92.mp3"
+      },
       "audioUrl": "https://www.soundhelix.com/examples/mp3/SoundHelix-Song-7.mp3",
       "title": "Argwan",
       "duration": 42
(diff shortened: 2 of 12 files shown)


aabounegm:
Houston, we have a problem.
In the service worker (from the template given by SvelteKit), there was this line:
https://github.com/aabounegm/cast/blob/209831364f2ad174cbd608f8cf5c72d62959b12c/src/service-worker.ts#L65
The `event.request.headers.has('range')` in particular is interesting for us. This header is used to ask the server for only a part of the file if it can. It is typically used when streaming audio/video so that it is not blocked until the whole file arrives, and also doesn't send a 1-hour file when a user only wants to listen to 5 minutes. It is also helpful when seeking to a direct position in a file, so it doesn't have to load everything before it.
However, for historical reasons, some browsers (up to very recently) had problems with that header in service workers that caused code as simple as:
```js
self.addEventListener('fetch', (event) => {
  event.respondWith(fetch(event.request));
});
```
to not work properly. In short, it drops that `Range` header so it doesn't reach the server. This is the reason the guard above was added to the service worker, so it would simply ignore these requests and let the browser handle them as usual.

Now, even if we decide we only support **Chrome v87**+, we would still have the issue of not properly handling the `Range`, both in the request and the response. That is because if the request asked only for a part of a file and we cached it, we might get a cache hit in the future for the same URL but with a different range, but we'd still return the original one we cached. And even if we cached the entire file, we would still return it all entirely when only a `range` of it is requested. It is possible to handle this in the service worker code, but it is not simple.

Currently, I went around it by allowing the service worker to handle requests with the `Range` header, completely overlooking the problems mentioned above.

More details about the problem can be found in [this web.dev post](https://web.dev/sw-range-requests/), which also links to a solution that uses a Workbox plugin ([`workbox-range-request`](https://developer.chrome.com/docs/workbox/modules/workbox-range-request/)). We will probably not use Workbox (I'm not a fan of it), but the tedious, manual solution can also be found in [this article](https://samdutton.github.io/samples/service-worker/prefetch-video/).

Even if we decide to ignore the problem for now, we should open an issue with the content of this comment. Should we ignore it, or should I try to implement the ugly fix above?

aabounegm:
As for why I skipped the test after all of that, that's because the `cy.expectPlayingAudio()` command isn't working as it should, and I already spent upwards of 5 hours on this task alone. At the very least, I need a little break)

aabounegm:
After wasting more time and effort, it became obvious that it's just not worth it. The tests should theoretically work, but there are bugs from Cypress and Chrome and even Supabase that prevent it from passing. I'm not willing to spend a few more hours coming up with workarounds.
On the plus side, it did help find a few major bugs.

src/lib/pages/podcast-gallery/api/load-podcasts.ts
```diff
@@ -5,7 +5,8 @@ export const loadPodcasts: Load = async () => {
   try {
     podcasts.set(await podcastList());
     return {};
-  } catch {
+  } catch (e) {
+    console.error(e);
```
illright, line 9: Issue: we're exposing stack traces here, let's not do that

illright (changes requested):

src/lib/pages/podcast-gallery/api/load-podcasts.ts
aabounegm: Why not? What's wrong with exposing the stack trace of a frontend app that's open-source?
And for debugging purposes, at least enable logging during dev

illright: OWASP thinks it's a security vulnerability. I don't agree with it either, but for the purposes of this project it's very much to our advantage to conform

illright (approved):

PR #207, opened 2022-04-20
Feature: audio error reporting

### LINKED ISSUE #124: Add error handing to the playing and seeking functionality
Blocked by #107 

If playing or seeking raises an exception (for example, when the connection is unstable), we should let the user know what's going on.

khaledismaeel:
Closes #124.

The implementation contains a dirty workaround. For some reason, the error code set by the `HTMLAudioElement` when the user is offline is _not_ `MEDIA_ERR_NETWORK`, but rather `MEDIA_ERR_SRC_NOT_SUPPORTED`, which causes a improper error message to appear. I tried to trace the issue around for a little, but I didn't get anywhere (got lost around MDN). I therefore dealt with the case of offline users separately.


src/lib/entities/audio/model/audio-instance.ts
```diff
@@ -42,3 +43,29 @@ function clampTime(time: number) {
   const d = get(duration);
   return Math.max(0, Math.min(time, isNaN(d) ? 0 : d));
 }
+
+export function reportError(el: HTMLElement) {
+  // shameful workaround
+  if (!window.navigator.onLine) {
+    snackbar({ text: 'You are offline. Please connect to the internet and try again.' });
+    return;
+  }
```
aabounegm, line 52: I think this should be moved after the `if (err == null) return` line, because why would we display it if there are no errors?

src/lib/entities/audio/model/audio-instance.ts
```diff
@@ -42,3 +43,29 @@ function clampTime(time: number) {
...
+
+  const audioEl = el as HTMLAudioElement;
```
aabounegm, line 54: A comment documenting the reason for the type assertion instead of using it as the parameter type would be helpful

src/lib/entities/audio/model/audio-instance.ts
aabounegm: Also, it isn't so shameful)

aabounegm (approved):

PR #215, opened 2022-04-28
No layout shifts

### LINKED ISSUE #212: Prevent layout shifts on the home page
We should add a skeleton for the "Trending" podcast shelf to avoid nasty layout shifts

### LINKED ISSUE #216: Server-side render the listening history
We will have to shift the storage of listening history to cookies instead of local storage

illright:
Closes #212 and closes #216. But I can't tell if the performance is actually worse now, Lighthouse is being kinda weird. The app certainly feels better now, so idk


illright:
Or maybe it feels worse

illright:
No, it's actually better, but there's still a layout shift for when there's no listening history

aabounegm (approved):

illright:
@aabounegm can you handle the Applitools differences?

aabounegm:
I did, but it doesn't update the workflow status for the integration tests.

Leaving this link here for the future, but probably even this wouldn't update the original workflow status: https://applitools.com/blog/link-github-actions/
Perhaps we can combine the link above with a `cy.on('fail', console.error)` inside the relevant test

VanishMax:
@illright Why won't we just remove the "Recently listened" when there's no podcasts?
Or, actually, we could keep track of listened podcasts in the cookies instead of LocalStorage, so this data is fulfilled on the backend

illright:
You're right, actually, storing the history in cookies might be a great idea to optimize this

aabounegm:
The link was about a dedicated workflow for checking the Applitools status. As I said earlier, approving the changes in Applitools' dashboard does **not** change Cypress' check status into a green checkmark. I'm also not sure if that link helps with this issue.

But also, I think the issue here makes sense. There might not be large content shifts anymore, but there is a header with empty content. I believe the behavior on the left is still desirable.

image

VanishMax (approved):
Great work!

PR #222, opened 2022-05-02
Prevent syncing when the user is not signed in

### LINKED ISSUE #221: The app tries to sync history even when the user is not signed in

illright:
Closes #221

### DIFF
--- a/src/lib/features/listening-history/model/store.ts
+++ b/src/lib/features/listening-history/model/store.ts
@@ -1,8 +1,12 @@
+import { get } from 'svelte/store';
 import { persistentWritable } from 'svelte-persistent-writable';
+import { dev } from '$app/env';
+import type { PostgrestResponse } from '@supabase/supabase-js';
 
 import { currentlyPlayingEpisode } from '$lib/entities/episode';
 import { user } from '$lib/entities/user';
 import type { Episode } from '$lib/shared/api';
+import { snackbar } from '$lib/shared/ui/snackbar';
 
 import { fetchHistory } from '../api/fetch-history';
 import { addToCloudListeningHistory } from '../api/history-table';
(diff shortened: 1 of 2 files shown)


alkaitagi (approved):

src/lib/features/listening-history/model/store.ts
```diff
@@ -31,12 +35,23 @@ const addToLocalListeningHistory = (podcastId: number) => {
...
+    } catch (error) {
+      if (dev) {
+        console.error(error);
+      }
+
+      if ((error as PostgrestResponse<unknown>).status >= 500) {
+        snackbar({ text: 'Failed to synchronize history due to a problem on our side. Sorry!' });
+      } else {
+        snackbar({
+          text: 'Either you are trying to do something shifty or your sign-in broke. Try sigining out and then in again.',
```
aabounegm, line 51: _sig**~~i~~**ning_

src/lib/features/listening-history/model/store.ts
```diff
@@ -31,12 +35,23 @@ const addToLocalListeningHistory = (podcastId: number) => {
 };
 
 export const addToListeningHistory = async (episode: Episode) => {
```
aabounegm, line 37: Seeing as there is no `await` here and the function doesn't return any promises (or even values), why is it marked `async`?

src/lib/features/listening-history/model/store.ts
```diff
@@ -31,12 +35,23 @@ const addToLocalListeningHistory = (podcastId: number) => {
...
-  const id = episode.podcastID;
-  addToLocalListeningHistory(id);
-  try {
-    addToCloudListeningHistory(id);
-  } catch (e) {
-    alert((e as Error).message);
+  addToLocalListeningHistory(episode.podcastID);
+  if (get(user) !== null && window.navigator.onLine) {
```
aabounegm, line 39: The nesting in this function is a little extra. How about this gets converted into a guard statement?
```suggestion
  if (get(user) === null || !window.navigator.onLine) return;
```

src/lib/features/listening-history/model/store.ts
aabounegm: I see that an async function (`addToCloudListeningHistory`) is called inside, but will the `try`/`catch` actually take effect without `await` or will it be a normal promise rejection?
Perhaps unit tests would be useful here

illright: I'll add unit tests (maybe) later

src/lib/features/listening-history/model/store.ts
illright: I don't like them. They're not expressive, they read like a machine, not like a human

src/lib/features/listening-history/model/store.ts
aabounegm: Ok. And if you are interested in the answer to my other questions, [this SO answer](https://stackoverflow.com/questions/30649994/can-i-catch-an-error-from-async-without-using-await) says that un`await`ed promise rejections will **not** trigger the `catch` block.

PR #228, opened 2022-05-06
Episode likes E2E tests

### LINKED ISSUE #196: Create E2E tests related to episode likes
See https://github.com/aabounegm/cast/issues/176#issuecomment-1088375282 for a list of requirements, section **Episode likes**.

**Acceptance criteria**:

aabounegm:
Closes #196 

The first requirement was already satisfied before. The second one has a problem due to re-rendering caused by mutating a list, as outlined in the `TODO` comment there.


cypress/integration/like.spec.ts
```diff
@@ -30,3 +30,91 @@ it('likes an episode then finds it in the library', () => {
...
+      .select('id, episodes (title)')
+      .limit(1)
+      .single();
+    expect(podcast).to.be.not.null;
+    if (podcast === null) return;
+    const titles = podcast.episodes.map(({ title }) => title);
+    cy.visitAndWaitForHydration(`/podcasts/${podcast.id}`);
+
+    // Like all the episodes
+    cy.findAllByTestId('episode-card').each((card, index) => {
```
illright, line 47: suggestion: let's ditch the test IDs
```suggestion
    cy.findAllByRole('article').each((card, index) => {
```

cypress/integration/like.spec.ts
```diff
@@ -30,3 +30,91 @@ it('likes an episode then finds it in the library', () => {
...
+      cy.wrap(card).within(() => {
+        cy.findByText(titles[index]).should('exist');
```
illright, line 49: issue: this doesn't really test that this title is linked as an accessible name of the episode card

suggestion: I think the test would be better if we were to loop over the titles and for each title `findByRole('article', { name: title })`. That way we test that each episode has an accessible name, and also it allows us to immediately chain a `.within()` off of it

cypress/integration/like.spec.ts
```diff
@@ -30,3 +30,91 @@ it('likes an episode then finds it in the library', () => {
...
+      cy.wrap(card).within(() => {
+        cy.findByRole('button', {
+          name: 'Remove the like',
+        }).click();
+      });
+    });
+
+    // Expect the episode titles to not be in the library
+    titles.forEach((title) => {
+      cy.findByLabelText(/favorites/i)
```
illright, line 74: issue: I'm not sure this is the best query to use here

suggestion:
```suggestion
      cy.findByRole('region', { name: /favorites/i })
```

cypress/integration/like.spec.ts
```diff
@@ -30,3 +30,91 @@ it('likes an episode then finds it in the library', () => {
...
+      .from<SBPodcast>('podcasts')
+      .select('id, episodes (title)')
+      .limit(1)
+      .single();
+    expect(podcast).to.be.not.null;
+    if (podcast === null) return;
+    const titles = podcast.episodes.map(({ title }) => title);
+    cy.visitAndWaitForHydration(`/podcasts/${podcast.id}`);
+
+    // Like all the episodes
```
illright, line 93: question: why is it necessary to like all episodes? I think the test would run faster if we just stuck to one

illright (changes requested):
I'd prefer if you moved the tests related to episode likes into `integration/requirements/episode-likes/`, possibly removing the `like.spec.ts`

cypress/integration/like.spec.ts
aabounegm: right, and it would also solve (sweep under the carpet) the issue with the re-rendering since we would only check for one card, but I didn't know based on what to pick an episode (always the first one? at random?) and what to do if there were no episodes yet

illright: I'd pick the first episode and rely on the fact that every podcast has at least one episode. Check my other tests to see how I've done it

cypress/integration/like.spec.ts
aabounegm: This is inside `cy.wrap(card)`, which is already the `<article>`. I'm just verifying that it has the text inside the card. Also, `{ name: title }` would check for an accessible [`title` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title), not text content.

illright: Yeah, but it's not enough to have text inside the card for assistive technologies to recognize that this text is the accessible name of the entire card. I've implemented this using `aria-labelledby`.

`{ name: title }` would first check for an accessible name coming from `aria-labelledby`, then `aria-label`, and only then the `title`

aabounegm: Ah right, it was failing only because it's inside the `within`. Actually, now that you point it out, I think that "_to recognize that this text is the accessible name of the entire card_" is not the responsibility of this test and this line should probably be removed altogether, right?

illright: Yeah, perhaps it's not

cypress/integration/requirements/episode-likes/persist-like.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+          name: 'Like this episode',
+        }).click();
+      });
+
+    cy.visitAndWaitForHydration('/library');
+
+    // Expect the liked episode title to be in the library
+    cy.findByLabelText(/favorites/i).contains(episodeTitle);
+
+    // All of the above is basically "likes get listed in the library" test
```
illright, line 32: suggestion (non-blocking): maybe it's best to just combine the two tests into one then?

illright (changes requested):

cypress/integration/requirements/episode-likes/like-in-library-and-persists.spec.ts
```diff
@@ -15,26 +15,25 @@ it('persists liked episodes even if not authenticated', () => {
...
+    cy.findByLabelText(/favorites/i).within(() => {
+      cy.findByRole('article', { name: episodeTitle });
+    });
 
     // Reload the library page
     cy.visitAndWaitForHydration('/library');
 
     // Expect the episode title to still be in the library
-    cy.findByLabelText(/favorites/i).contains(episodeTitle);
+    cy.findByLabelText(/favorites/i).within(() => {
```
illright, line 35:
```suggestion
    cy.findByRole('region', { name: /favorites/i }).within(() => {
```

cypress/integration/requirements/episode-likes/like-in-library-and-persists.spec.ts
```diff
@@ -15,26 +15,25 @@ it('persists liked episodes even if not authenticated', () => {
...
+    });
 
     cy.visitAndWaitForHydration('/library');
 
-    // Expect the liked episode title to be in the library
-    cy.findByLabelText(/favorites/i).contains(episodeTitle);
-
-    // All of the above is basically "likes get listed in the library" test
+    // Expect the episode title to be in the library
+    cy.findByLabelText(/favorites/i).within(() => {
```
illright, line 27:
```suggestion
    cy.findByRole('region', { name: /favorites/i }).within(() => {
```

illright (approved):

PR #229, opened 2022-05-07
E2E-test listening

### LINKED ISSUE #180: Create end-to-end tests related to audio listening
See https://github.com/aabounegm/cast/issues/176#issuecomment-1088375282 for a list of requirements, section **Audio listening** (4 requirements)

**Acceptance criteria**:

illright:
Closes #180

I couldn't for the life of me figure out how to test the scrubbing bar. Apparently, testing those is not a thing in Cypress either. Screw it


cypress/support/expect-playing-audio.ts
```diff
@@ -10,4 +10,14 @@ const expectPlayingAudio = () => {
   });
 };
 
-export default expectPlayingAudio;
+export const expectNoPlayingAudio = () => {
```
aabounegm, line 13: Instead of duplicating the entire function and changing just one word, couldn't the common functionality be factored out to another function?

package.json
```diff
@@ -78,6 +79,7 @@
     "eslint-plugin-jest-dom": "^4.0.1",
     "eslint-plugin-svelte3": "^3.4.1",
     "eslint-plugin-testing-library": "^5.2.0",
+    "html-to-text": "^8.2.0",
```
aabounegm, line 82: Don't we already have `dompurify`? How is this different?

src/lib/widgets/bottom-bar/ui/bottom-bar.svelte
```diff
@@ -29,7 +29,7 @@
 >
   <BottomBarNavItem name="Gallery" icon={IconVideoLibrary} href="/" showLabel={!hasAudio} />
   {#if hasAudio}
-    <MiniPlayer on:click={() => dispatch('miniplayer-click')} />
+    <MiniPlayer on:expand={() => dispatch('miniplayer-click')} />
```
aabounegm, line 32: Why did you rename one event but not the other? (referring to miniplayer-**click** event)

cypress/support/expect-playing-audio.ts
illright: I tried playing with it but it just didn't work well in any way, so I decided to just let it go. Suggestions welcome

package.json
illright: DOMPurify only strips dangerous tags, afaik

src/lib/widgets/bottom-bar/ui/bottom-bar.svelte
illright: Couldn't be bothered refactoring more than what was necessary. Also to keep the diff clean. I don't think this is a problem

PR #241, opened 2022-05-08
Clear likes and listening history on sign out

### LINKED ISSUE #237: Clear local likes and history upon signout
See #208, tour 2, step 2

aabounegm:
Closes #237

### DIFF
--- a/src/lib/features/authenticate/model/sign-out.ts
+++ b/src/lib/features/authenticate/model/sign-out.ts
@@ -1,9 +1,5 @@
 import { supabaseClient } from '$lib/shared/api';
-import { likesStore } from '$lib/features/like-episode';
-import { listeningHistory } from '$lib/features/listening-history';
 
 export async function signOut() {
   await supabaseClient.auth.signOut();
-  likesStore.set([]);
-  listeningHistory.set([]);
 }
--- a/src/lib/features/authenticate/model/sign-out.ts
+++ b/src/lib/features/authenticate/model/sign-out.ts
@@ -1,5 +1,9 @@
 import { supabaseClient } from '$lib/shared/api';
+import { likesStore } from '$lib/features/like-episode';
+import { listeningHistory } from '$lib/features/listening-history';
 
-export function signOut() {
-  supabaseClient.auth.signOut();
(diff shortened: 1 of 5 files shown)


src/lib/features/authenticate/model/sign-out.ts
```diff
@@ -1,5 +1,9 @@
 import { supabaseClient } from '$lib/shared/api';
+import { likesStore } from '$lib/features/like-episode';
+import { listeningHistory } from '$lib/features/listening-history';
```
illright, line 3: issue: you have been smitted by the gods of Feature-Sliced Design. -10 FSD coins in your afterlife. Can't import features from features

suggestion: we already have a `user` store subscription in both of these features, right now we only take action if `$user` is truthy. In the case of signout, it's just a matter of adding an `else` to that `if`
https://github.com/aabounegm/cast/blob/bd8fc73fbd98ddfbea73112d6fdbc42413c7d46a/src/lib/features/like-episode/model/like.ts#L13-L18

illright (changes requested):

src/lib/features/authenticate/model/sign-out.ts
aabounegm: Yeah, I was kinda hoping you wouldn't notice that)
But I don't think your solution would work because this event is supposed to fire on page load, even though it currently doesn't due to a bug (according to an open issue on Supabase's repo), and that's also the behavior I'm familiar with from Firebase.
If we replaced this `user.subscribe` with `supabaseClient.auth.onAuthStateChange`, I could check if `event === 'SIGNED_OUT'` in the body, but as it is right now I just cannot depend on `user == null` to mean the action of sign out just took place. It can simply be the page load and `supabaseClient.auth.user()` is `null` because the user is not logged in. This would clear the history on every page load.

illright: > If we replaced this `user.subscribe` with `supabaseClient.auth.onAuthStateChange`

Let's try that then

illright (approved):
Oh, right, that's a nice way to do it :p

PR #242, opened 2022-05-08
Add local likes & history to cloud after signing in

### LINKED ISSUE #235: Upload likes and listening history upon connecting to the internet
See #208, tour 1, step 8. Apparently, we don't upload the locally stored likes and history right after signing in

aabounegm:
Closes #235

### DIFF
--- a/src/lib/features/authenticate/model/sign-in-snackbar.ts
+++ b/src/lib/features/authenticate/model/sign-in-snackbar.ts
@@ -1,5 +1,8 @@
 import { snackbar } from '$lib/shared/ui/snackbar';
 import { supabaseClient } from '$lib/shared/api';
+import { likesStore, addCloudLike } from '$lib/features/like-episode';
+import { listeningHistory, addToCloudListeningHistory } from '$lib/features/listening-history';
+import { get } from 'svelte/store';
 
 export function signInSnackbar() {
   const url = new URL(window.location.href);
@@ -14,5 +17,7 @@ export function signInSnackbar() {
     });
     url.searchParams.delete('snackbar');
     window.history.replaceState(null, '', url.toString());
+    get(likesStore).forEach(addCloudLike);
+    get(listeningHistory).forEach(addToCloudListeningHistory);
   }
 }
--- a/src/lib/features/like-episode/index.ts
+++ b/src/lib/features/like-episode/index.ts
@@ -1,3 +1,4 @@
 export { default as LikeButton } from './ui/like-button.svelte';
 export { likesStore, toggleLike } from './model/like';
 export { populateLikes } from './model/populate-likes';
+export { addCloudLike } from './api/favourites-table';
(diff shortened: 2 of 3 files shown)


src/lib/features/authenticate/model/sign-in-snackbar.ts
```diff
@@ -1,5 +1,8 @@
 import { snackbar } from '$lib/shared/ui/snackbar';
 import { supabaseClient } from '$lib/shared/api';
+import { likesStore, addCloudLike } from '$lib/features/like-episode';
+import { listeningHistory, addToCloudListeningHistory } from '$lib/features/listening-history';
```
illright, line 4: issue: FSD, cross-imports of features

src/lib/features/authenticate/model/sign-in-snackbar.ts
```diff
@@ -14,5 +17,7 @@ export function signInSnackbar() {
     });
     url.searchParams.delete('snackbar');
     window.history.replaceState(null, '', url.toString());
+    get(likesStore).forEach(addCloudLike);
```
illright, line 20: question: why did you decide to do it here, of all the places?)

suggestion: the same as in that other PR, let's do it here:

https://github.com/aabounegm/cast/blob/bd8fc73fbd98ddfbea73112d6fdbc42413c7d46a/src/lib/features/like-episode/model/like.ts#L13-L18

illright (changes requested):

src/lib/features/authenticate/model/sign-in-snackbar.ts
aabounegm: Because I didn't want to add the like to Supabase every time the user loads the page (I'm not sure if there are any uniqueness constraints set in the backend). This seemed like the only function where I guarantee the user has just signed in after being logged out before and having added some episode to the favorites.
I agree it's not the best solution, especially that it only works if the user signed in through the snackbar that appears after liking an episode unauthenticated, but I also couldn't think of another quick solution that doesn't involve querying the backend likes/history and diffing the data.

src/lib/features/authenticate/model/sign-in-snackbar.ts
aabounegm: Same as https://github.com/aabounegm/cast/pull/241#discussion_r867538147

src/lib/features/authenticate/model/sign-in-snackbar.ts
illright: There are uniqueness constraints on the data, so uploading a duplicate like/history won't really change anything except for the `created_at` timestamp, which we don't use anyway, except for ordering.

Technically, we could have another `derived` store that tracks the previous value of the `user` store, and expose to the public yet another `derived` store that provides access to them both, however, that is slightly janky, so proceed at will

aabounegm: Still, I would prefer not to depend on constraints in the backend, especially that the function name literally says "add". It's better to validate the data on the client side first and consider the backend validation as a safety measure

illright: Alright, then diffing on the client it is. Shouldn't be too expensive/hard

PR #244, opened 2022-05-09
Remove the code that was pausing playback while scrubbing

### LINKED ISSUE #231: Fix the scrubbing bar pausing playback in Firefox
As Khaled has discovered in #208, in Firefox, when you scrub, the playback stops

illright:
Closes #231 

We previously wrote some code that would pause the playback while dragging the scrubbing bar, but turns out it's a problem for Firefox who dispatches yet another `input` event after the `pointerup` event, which is what was causing the pause.

I tried all sorts of solutions, all sorts of state tracking, some of that would fix the issue but introduce a worse one in one of the browsers. This behaviour seems to be the most stable and harmless. Moreover, Firefox doesn't play while you scrub, only Chrome does, but it's not a problem either way

### DIFF
--- a/src/lib/features/playback-controls/ui/scrubbing-bar.svelte
+++ b/src/lib/features/playback-controls/ui/scrubbing-bar.svelte
@@ -12,7 +12,7 @@
   $: remaining = duration - position;
 
   interface ScrubbingBarEvents {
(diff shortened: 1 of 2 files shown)


aabounegm (approved):
Didn't test, but LGTM

